Issue #4, opened 2022-11-07
Move day time calculation to the GameTimer


vE5li:
Currently the variable that controls the day night cycle is calculated inside the the main function. This functionality should be moved to the `GameTimer` and only expose a single method for getting the current value, plus one or more additional feature gated methods for setting the time in a debug build.

PR #9, opened 2022-11-15
Move timers to timer module

goalves:
Related to: https://github.com/vE5li/korangar/issues/4

Took the opportunity to move the `animation_timer` as well since it was only used and calculated in the main.rs file and was tightly coupled with the `GameTimer` module.

Also added some unit testing to the `GameTimer` module.


vE5li:
Great work 👍🏻

Issue #6, opened 2022-11-07
Improve client tick calculation


vE5li:
Currently there is a lot of desync betwenn the server and the client, which causes frequent stuttering and jumping of animations. The cause is my rather [crude implementation](https://github.com/vE5li/korangar/blob/165d3e21d88c39f168c187e29da58f996de195ed/src/system/timer.rs#L44) of the client tick calculation. rAthena's implementation of the client timer can be found [here](https://github.com/rathena/rathena/blob/774d4ca2e8119e346ab502c6f76b0d2c1e9bcddd/src/common/timer.cpp#L139) and should hopefully be all the reference needed for this ticket.

PR #11, opened 2022-11-28
Stop updating game-tick with server-tick

goalves:
This caused SOME of the instability and "rubber banding", especially with animations that are entirely client-side. 

This doesn't resolve all issues described in #6 since rubber banding between server-client can still happen for several other reasons. 

Leaving it as a draft still because I am still trying to compare if the animation speed is correct... it looks better compared to the actual client now but I am unsure if these changes won't affect anything else:

```
self.client_tick.0 += (delta_time * 1000.0) as u32;
```

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -497,8 +497,9 @@ fn main() {
                             // cursor always look correct.
                             interface.set_start_time(game_timer.get_client_tick());
                         }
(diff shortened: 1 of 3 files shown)


vE5li:
This is fixed as of https://github.com/vE5li/korangar/commit/6f48305b0b89fbb83d7b363d4eaadc0a3510a67a

PR #12, opened 2022-11-28
Allow colors by default on windows (only disable terminal colors with plain feature)

goalves:
Since PowerShell + windows terminal supports the same colors as UNIX terminals we can have it as default and use the plain feature to remove the colors when needed.

This wasn't discussed before but I was testing some things on windows and noticed that colors were missing since the colors feature was disabled by default on windows. What do you think about it? @vE5li

### DIFF
--- a/src/debug/logging/colors.rs
+++ b/src/debug/logging/colors.rs
@@ -1,4 +1,4 @@
-#[cfg(any(target_os = "windows", feature = "plain"))]
+#[cfg(feature = "plain")]
 mod colors {
     pub const GREEN: &str = "";
     pub const RED: &str = "";
@@ -8,7 +8,7 @@ mod colors {
     pub const NONE: &str = "";
 }
 
-#[cfg(not(any(target_os = "windows", feature = "plain")))]
+#[cfg(not(feature = "plain"))]
 mod colors {
     pub const GREEN: &str = "\x1B[32m";
     pub const RED: &str = "\x1B[31m";


vE5li:
Sounds very good 👍🏻
I'd be down to merge it if you want to remove the draft status

vE5li:
Ater you removed the `any`s from thr cfg though :D

src/debug/logging/colors.rs
```diff
@@ -8,7 +8,7 @@ mod colors {
     pub const NONE: &str = "";
 }
 
-#[cfg(not(any(target_os = "windows", feature = "plain")))]
+#[cfg(not(any(feature = "plain")))]
```
vE5li, line 11:
```suggestion
#[cfg(not(feature = "plain"))]
```

src/debug/logging/colors.rs
```diff
@@ -1,4 +1,4 @@
-#[cfg(any(target_os = "windows", feature = "plain"))]
+#[cfg(any(feature = "plain"))]
```
vE5li, line 1:
```suggestion
#[cfg(feature = "plain")]
```

vE5li (changes requested):

src/debug/logging/colors.rs
goalves: Ops, completely forgot to remove the `any`

goalves:
Updated to remove the `any`s @vE5li

PR #13, opened 2022-12-02
Fix unexpected packet parsing error while logging in

### LINKED ISSUE #10: Selecting a character sometimes crashes the client because of an unexpected packet
On occasion you will get this crash when selecting a character: `thread 'main' panicked at 'called Result::unwrap() on an Err value: "invalid signature 0x1d 0x0b"'`.

This is most likely happening because the [select character method](https://github.com/vE5li/korangar/blob/main/src/network/mod.rs#L2310) is waiting for certain packets to arrive in a specific order. If any packet is sent out of order or we receive an additional packet, the function will panic.

A possible solution would be to not wait for any packets in this method and instead let the main loop handle the response packets, ~~which would require adding some member variables to the network system to keep track of the current state~~.

Thank you to @goalves for correcting this bug report.

goalves:
Closes #10

Didn't really need to store any state elsewhere because I managed to leverage the `CharacterSelectionSuccessPacket` to load the correct map.

Ended up "cheating" and setting that the user is at 0,0 on the map while loading (because we will soon receive other packets with the user's correct position. Is that ok?

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -29,7 +29,7 @@ use std::io::Cursor;
 use std::rc::Rc;
 use std::sync::Arc;
 
-use cgmath::Vector2;
+use cgmath::{Vector2, Zero};
 use image::io::Reader as ImageReader;
 use image::{EncodableLayout, ImageFormat};
 use procedural::debug_condition;
@@ -614,7 +614,7 @@ fn main() {
                                         &script_loader,
                                         &map,
                                         character_information,
-                                        Vector2::new(0, 0),
+                                        Vector2::zero(),
                                         client_tick,
                                     );
                                     let player = Entity::Player(player);
--- a/src/main.rs
+++ b/src/main.rs
@@ -29,6 +29,7 @@ use std::io::Cursor;
 use std::rc::Rc;
 use std::sync::Arc;
 
+use cgmath::Vector2;
 use image::io::Reader as ImageReader;
 use image::{EncodableLayout, ImageFormat};
 use procedural::debug_condition;
@@ -600,14 +601,8 @@ fn main() {
                         UserEvent::ReloadTheme => interface.reload_theme(),
                         UserEvent::SaveTheme => interface.save_theme(),
                         UserEvent::SelectCharacter(character_slot) => {
-                            match networking_system.select_character(character_slot, &chat_messages) {
-                                Ok((map_name, player_position, character_information, client_tick)) => {
-                                    // TODO: this will do one unnecessary restore_focus. check if
-                                    // that will be problematic
-                                    interface.close_window_with_class(&mut focus_state, CharacterSelectionWindow::WINDOW_CLASS);
-                                    interface.open_window(&mut focus_state, &CharacterOverviewWindow::new());
-                                    interface.open_window(&mut focus_state, &ChatWindow::new(chat_messages.clone(), font_loader.clone()));
-
+                            match networking_system.select_character(character_slot) {
+                                Ok((character_information, map_name)) => {
                                     map = map_loader
                                         .get(map_name, &mut game_file_loader, &mut model_loader, &mut texture_loader)
                                         .unwrap();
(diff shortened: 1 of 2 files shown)


vE5li:
I really like what you are doing here but I'm wondering if this is actually the bug that is described in #10, since this is changing the character selection and not the log in.

vE5li:
Apart from that, I think setting the position to 0,0 is not ideal but considering how much code you were able to remove I think it's a fair trade off. Did you test to see if this is affecting the camera while loading the map? Because usually the camera is focused on the player, so the camera might be at 0,0 for a very brief moment before focusing on the correct position, which I would like to avoid.

src/main.rs
```diff
@@ -619,14 +614,20 @@ fn main() {
                                         &script_loader,
                                         &map,
                                         character_information,
-                                        player_position,
+                                        Vector2::new(0, 0),
                                         client_tick,
                                     );
                                     let player = Entity::Player(player);
 
                                     player_camera.set_focus_point(player.get_position());
```
vE5li, line 622: This line specifically is already making me a bit worried.

vE5li:
Also why did you move the window opening and closing part to below the map loading? I definitely don't mind as it does make a bit more sense, just curious.

goalves:
> I really like what you are doing here but I'm wondering if this is actually the bug that is described in #10, since this is changing the character selection and not the log in.

I actually only had this issue (and reproduced it properly adding some waits/etc) in the character selection process, never in the log_in function. 🤔 I think the log_in piece doesn't really fetch/use data from the map server. 

> Apart from that, I think setting the position to 0,0 is not ideal but considering how much code you were able to remove I think it's a fair trade off. Did you test to see if this is affecting the camera while loading the map? Because usually the camera is focused on the player, so the camera might be at 0,0 for a very brief moment before focusing on the correct position, which I would like to avoid.

I was also worried about that but couldn't really find a better solution to make it better without really changing lots of code... Currently, the screen does "flash" in at (0, 0) and then moves to the correct place. I can try to store the map/char information locally and once we receive the first `MapChangePacket` we create the entity with the correct position, would that be better?

> Also why did you move the window opening and closing part to below the map loading? I definitely don't mind as it does make a bit more sense, just curious :)

I think I was trying to make the flow make sense in my head and moved it down below (load map, create the entity, open/close UI windows, etc), definitely wasn't on purpose 🤦. Do you want me to change it back?

src/main.rs
goalves: Kind of answered in the general comment but the screen does flash at 0,0 and the moves the player to the correct position. I proposed a possible solution to it but I am open to other suggestions on how to make it better

vE5li:
> I actually only had this issue (and reproduced it properly adding some waits/etc) in the character selection process, never in the log_in function. thinking I think the log_in piece doesn't really fetch/use data from the map server.

Ah fair enough. I guess I messed up the ticket, I will fix the description later.
> I was also worried about that but couldn't really find a better solution to make it better without really changing lots of code... Currently, the screen does "flash" in at (0, 0) and then moves to the correct place. I can try to store the map/char information locally and once we receive the first MapChangePacket we create the entity with the correct position, would that be better?

I think the best solution heavily depends on if the client should have loading screens because in that case the camera position will be irrelevant until everything is loaded. Since I'm currently more leaning towards adding loading screens I would say let's accept this little visual bug, we can always fix it later if necessary.

> I think I was trying to make the flow make sense in my head and moved it down below (load map, create the entity, open/close UI windows, etc), definitely wasn't on purpose facepalm. Do you want me to change it back?

No not at all, I actually like it better this way :)

src/main.rs
```diff
@@ -619,14 +614,20 @@ fn main() {
                                         &script_loader,
                                         &map,
                                         character_information,
-                                        player_position,
+                                        Vector2::new(0, 0),
```
vE5li, line 617: Minor comment, maybe use Vector2::zero()

goalves: Didn't know it existed, thanks!👍

src/network/mod.rs
```diff
@@ -2658,6 +2625,10 @@ impl NetworkingSystem {
                             equipped_position: EquipPosition::None,
                         });
                     }
+                } else if let Ok(_) = Packet8302::try_from_bytes(&mut byte_stream) {
```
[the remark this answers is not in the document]
goalves: I can actually change it to `_packet` if you prefer (since that is how the code currently is)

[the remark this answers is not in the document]
goalves: Cool, will change all packet discards to `_`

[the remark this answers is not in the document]
goalves: Done on [26404e8](https://github.com/vE5li/korangar/pull/13/commits/26404e85b1086e2bd31b3e9340e7839a3488cfaa)

vE5li:
Very nice cleanup! :+1:

PR #14, opened 2022-12-15
	Map loader using ByteConvertable trait

### LINKED ISSUE #5: Transition map loader to use the ByteConvertable trait
The map loader is the last loader that does not utilize the `ByteConvertable` trait for loading. This ticket is more involved but at the same time relatively straightforward, as there is a lot of reference in the other loaders. The [model loader](https://github.com/vE5li/korangar/blob/main/src/loaders/model/mod.rs) specifically should be very useful because it does very similar things.

I got started on this in the past and the [code](https://github.com/vE5li/korangar/blob/2ae74b60d32ddff73b7f1de4a69d9739d97bad55/src/loaders/map/mod.rs#L127) is still in the model loader.

Jlabarca:
WIP - Closes #5  

Using`ByteConvertable `macro when possible. 
Also extracted part of mod.rs into two new local modules:
- `map_data.rs` containing `ByteConvertable` structs
- `vertices.rs` for vertices and buffers related code


src/loaders/version.rs
```diff
@@ -1,28 +1,76 @@
...
+    pub major: u8,
+    pub minor: u8,
+}
+
+impl<T> From<Version<T>> for InternalVersion {
+    fn from(version: Version<T>) -> Self {
+        let Version {
+            major,
+            minor,
+            phantom_data: PhantomData,
```
vE5li, line 67: you can just put `..` here instead, since we don't care about the `PhantomData`

src/world/light/mod.rs
```diff
@@ -6,7 +6,7 @@ use crate::graphics::*;
 #[cfg(feature = "debug")]
 use crate::world::MarkerIdentifier;
 
-#[derive(PrototypeElement, PrototypeWindow, new)]
+#[derive(ByteConvertable, PrototypeElement, PrototypeWindow, new)]
```
vE5li, line 9: I assume this is a left over from you trying to derive `ByteConvertable` for `LightSource`?

src/world/object/mod.rs
```diff
@@ -12,11 +12,15 @@ use crate::world::*;
 pub struct Object {
     pub name: Option<String>,
     pub model_name: String,
-    pub model: Arc<Model>,
+    pub model: Option<Arc<Model>>,
     pub transform: Transform,
 }
 
 impl Object {
+    pub fn set_model(&mut self, model: Arc<Model>) {
```
vE5li, line 20: I really don't like the idea of having this be an option. I think given the design that you have right now you should create an extra struct called `ObjectData` that gets deserialized and then in the second step use that to create an actual `Object`

src/loaders/map/map_data.rs
```diff
@@ -0,0 +1,175 @@
...
+    pub texture_count: i32,
+    pub texture_name_length: i32,
+    #[repeating(self.texture_count)]
+    #[length_hint(self.texture_name_length)]
+    pub textures: Vec<String>,
+    pub light_map_count: i32,
+    pub light_map_width: i32,
+    pub light_map_height: i32,
+    pub light_map_cells_per_grid: i32,
+
```
vE5li, line 61: You should run `cargo fmt --all` to get rid of these empty lines

vE5li (commented):
I added a good amount of comments but don't be discouraged, it is a big MR after all. All in all I really like where this is going :+1:

src/loaders/map/map_data.rs
Jlabarca: not really, I ran fmt --all

src/loaders/map/resource.rs
```diff
@@ -17,3 +25,198 @@ impl ResourceType {
...
+            resources_amount,
+            objects,
+            light_sources,
+            sound_sources,
+            effect_sources,
+        }
+    }
+}
+
+#[derive(Debug, PrototypeElement, new)]
```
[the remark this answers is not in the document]
Jlabarca: To do that we should do something with this [line](https://github.com/vE5li/korangar/blob/735b18ec2a29d186e9030824b29f2b8e6c934e98/procedural/src/byte/convertable.rs#L32) 
having all fields as Optionals (because they have version rules) throws an error there, something like "`Concat not implemented for [()]`"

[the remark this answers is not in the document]
vE5li: Hmm fair enough, I will try to fix that

src/loaders/map/map_data.rs
vE5li: Yes that is my bad, I thought that `cargo fmt` removes these empty lines but apparently it does not. It would be great if you could remove them manually, alternatively I would also be fine with a comment describing the fields below instead of an empty line.

src/loaders/map/map_data.rs
```diff
@@ -0,0 +1,175 @@
...
+                self.lower_left_height,
+                f32::max(self.upper_left_height, self.upper_right_height),
+            ),
+        )
+    }
+}
+
+impl ByteConvertable for GroundTile {
+    fn from_bytes(byte_stream: &mut ByteStream, _: Option<usize>) -> Self {
+        let upper_left_height = byte_stream.float32();
```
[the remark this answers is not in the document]
Jlabarca: Got it, it could be a second issue to create and take after this one?
this pr is already too big

src/world/light/mod.rs
Jlabarca: LightSource is now ByteConvertable, so I removed the "new" trait

src/loaders/map/map_data.rs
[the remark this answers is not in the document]
vE5li: That sounds like a good idea :+1:

src/loaders/map/resource.rs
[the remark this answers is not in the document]
vE5li: It is now fixed on master

PR #20, opened 2023-07-19
Implement player job sprite

gustavobca:


eleriaqueen:
I checked branch out, I can't comment on the code, but it seems to work fine.

Screenshot_2023-07-20_14-19-44
Screenshot_2023-07-20_14-20-44
Screenshot_2023-07-20_14-21-19

vE5li:
I have some questions on this. First of all, how do other clients do this? Is this list always hard coded?
I am always hoping that we can get this information from the lua files in the `data/rdata.grf`.

Either way I think it's fine to hard code it for now, since having job sprites working is better than not having them work at all.

src/network/mod.rs
```diff
@@ -130,12 +130,24 @@ impl ChatMessage {
...
 #[derive(Copy, Clone, Debug, ByteConvertable, PrototypeElement)]
 pub enum Sex {
-    Male,
     Female,
+    Male,
     Both,
     Server,
 }
 
+impl PartialEq for Sex {
```
vE5li, line 139: This can be derived on the struct by adding `PartialEq` to the line
```rust
#[derive(Copy, Clone, Debug, ByteConvertable, PrototypeElement)]
```
It's good practice to also derive `Eq`. So basically that line would become
```rust
#[derive(Copy, Clone, Debug, PartialEq, Eq, ByteConvertable, PrototypeElement)]
```
And this manual implementation can be removed completely

src/world/entity/mod.rs
```diff
@@ -73,6 +79,93 @@ pub struct Common {
...
+            table.insert("KAGEROU", "KAGEROU");
+            table.insert("OBORO", "OBORO");
+
+            table
+        };
+    }
+
+    JOB_SPRITE_NAME_TABLE.get(player_job).copied().unwrap_or("ÃÊº¸ÀÚ")
+}
+
```
vE5li, line 168: If we do hard code this for now, I think it would be better to use a `match` instead of creating a `HashMap` and indexing it. So basically this code would become
```rust
match player_job {
    "NOVICE" => "ÃÊº¸ÀÚ",
    "SWORDMAN" => "°Ë»Ç",
    /* ... */
    _ => "ÃÊº¸ÀÚ",
}
```
Then again, it is possible that an index into a `HashMap` is faster than the `switch` statement and if you have good reason to believe that then I can be convinced to keep the hash map. Maybe you could use the new built-in profiler to test the performance of both solutions? :eyes:

src/world/entity/mod.rs
```diff
@@ -152,8 +257,19 @@ impl Common {
...
+        };
+
         let file_path = match entity_type {
-            EntityType::Player => format!("¸ó½ºÅÍ\\b_{}", script_loader.get_job_name_from_id(self.job_id)),
+            EntityType::Player => format!(
+                "ÀÎ°£Á·\\¸öÅë\\{}\\{}_{}",
+                sex_sprite_path,
+                get_sprite_path_for_player_job(&script_loader.get_job_name_from_id(self.job_id)),
+                sex_sprite_path
+            ),
```
vE5li, line 272: I know that this is 100% my fault but since you are adding more code to the functions could you move all the duplicate code of `new` and `reload_sprite` to a common function?

src/world/entity/mod.rs
```diff
@@ -591,6 +707,10 @@ impl Player {
         self.common.job_id = job_id;
     }
 
+    pub fn set_sex(&mut self, sex: Sex) {
+        self.common.sex = sex;
+    }
+
```
vE5li, line 713: I noticed that `set_sex` and `get_sex` are not used by anything, does it maybe make sense to remove them for now?

gustavobca:
> I have some questions on this. First of all, how do other clients do this? Is this list always hard coded? I am always hoping that we can get this information from the lua files in the `data/rdata.grf`.
> 
> Either way I think it's fine to hard code it for now, since having job sprites working is better than not having them work at all.

I don't know how other clients do this, and I didn't find any lua table with a reference to these sprites in the grf

src/world/entity/mod.rs
gustavobca: > If we do hard code this for now, I think it would be better to use a `match` instead of creating a `HashMap` and indexing it. So basically this code would become
> 
> ```rust
> match player_job {
>     "NOVICE" => "ÃÊº¸ÀÚ",
>     "SWORDMAN" => "°Ë»Ç",
>     /* ... */
>     _ => "ÃÊº¸ÀÚ",
> }
> ```
> 
> Then again, it is possible that an index into a `HashMap` is faster than the `switch` statement and if you have good reason to believe that then I can be convinced to keep the hash map. Maybe you could use the new built-in profiler to test the performance of both solutions? 👀

Using switch with job id should be better

eleriaqueen:
Have you tried looking into the files under `data/LuaFiles514/Lua Files/admin` ? Theres one that sounds promising called `pcjobname.lub` sadly my Wine acts up when I click on any lub file so I can't decrypt it to check :(

aleDsz:
> Have you tried looking into files in `data/LuaFiles514/Lua Files/admin` ? Theres one that sounds promising called `pcjobname.lub` sadly my Wine acts up when I click on any lub file so I can't decrypt it to check what's inside.

It uses a string name that isn't used to load sprites, because the Novice sprite uses another name than `NOVICE`:

```lua
PCJobNameTable = {
	[pcJobTbl.JT_NOVICE] = "NOVICE",
	-- [...]
}
```

eleriaqueen:
I've just finished doing a quick look inside a 2022 ragexe with and there seems to be an immense list of "*.spr"  strings there... Ghidra only seems to support utf8/16 and US ASCII so here's how it displays data 
```
        00d77c28 72              ??         72h    r
        00d77c29 6f              ??         6Fh    o
        00d77c2a 75              ??         75h    u
        00d77c2b 6c              ??         6Ch    l
        00d77c2c 6c              ??         6Ch    l
        00d77c2d 65              ??         65h    e
        00d77c2e 74              ??         74h    t
        00d77c2f 74              ??         74h    t
        00d77c30 65              ??         65h    e
        00d77c31 5c              ??         5Ch    \
        00d77c32 73              ??         73h    s
        00d77c33 74              ??         74h    t
        00d77c34 61              ??         61h    a
        00d77c35 67              ??         67h    g
        00d77c36 65              ??         65h    e
        00d77c37 41              ??         41h    A
        00d77c38 72              ??         72h    r
        00d77c39 72              ??         72h    r
        00d77c3a 6f              ??         6Fh    o
        00d77c3b 77              ??         77h    w
        00d77c3c 2e              ??         2Eh    .
        00d77c3d 73              ??         73h    s                                         ?  ->  00727073
        00d77c3e 70              ??         70h    p
        00d77c3f 72              ??         72h    r
```
"roulette\stagearrow.spr"

aleDsz:
> I've just finished doing a quick look inside a 2022 ragexe with and there seems to be an immense list of "*.spr" strings there... Ghidra only seems to support utf8/16 and US ASCII so here's how it displays data
> 
> ```
>         00d77c28 72              ??         72h    r
>         00d77c29 6f              ??         6Fh    o
>         00d77c2a 75              ??         75h    u
>         00d77c2b 6c              ??         6Ch    l
>         00d77c2c 6c              ??         6Ch    l
>         00d77c2d 65              ??         65h    e
>         00d77c2e 74              ??         74h    t
>         00d77c2f 74              ??         74h    t
>         00d77c30 65              ??         65h    e
>         00d77c31 5c              ??         5Ch    \
>         00d77c32 73              ??         73h    s
>         00d77c33 74              ??         74h    t
>         00d77c34 61              ??         61h    a
>         00d77c35 67              ??         67h    g
>         00d77c36 65              ??         65h    e
>         00d77c37 41              ??         41h    A
>         00d77c38 72              ??         72h    r
>         00d77c39 72              ??         72h    r
>         00d77c3a 6f              ??         6Fh    o
>         00d77c3b 77              ??         77h    w
>         00d77c3c 2e              ??         2Eh    .
>         00d77c3d 73              ??         73h    s                                         ?  ->  00727073
>         00d77c3e 70              ??         70h    p
>         00d77c3f 72              ??         72h    r
> ```
> 
> "roulette\stagearrow.spr"

I did the same before posting here, and I didn't found anything related to, sadly :/

eleriaqueen:
Hey, I dug deeper I can confirm class names are in the ragexe.
For instance, let's say Swordsman:
```
ISO-8859-1: °Ë»ç"
EUC-KR: 검사
Bytes: B0 CB BB E7
```
Ghidra: Found match:
```
        00d38674 b0              ??         B0h
        00d38675 cb              ??         CBh  
        00d38676 bb              ??         BBh
        00d38677 e7              ??         E7h
```
And we can see that the next one right near that address is 
```
        00d3867c b8              ??         B8h
        00d3867d b6              ??         B6h
        00d3867e b9              ??         B9h
        00d3867f fd              ??         FDh 
        00d38680 bb              ??         BBh
        00d38681 e7              ??         E7h
```
Which is Magician
```
ISO-8859-1: ¸¶¹ý»ç
EUC-KR: 마법사
Bytes: B8 B6 B9 FD BB E7
```
And the list continues, I saw Assassin a few strings further:
```
¾î¼¼½Å
어세신
BE EE BC BC BD C5
```

vE5li:
That is some very valuable insight :+1:
With that I feel comfortable merging this

PR #22, opened 2023-07-19
Only add Choice Button "cancel" if there is no choice cancel

gustavobca:


vE5li:
No complaints about the code. Do you know by any chance if other clients to this the same way?

gustavobca:
> No complaints about the code. Do you know by any chance if other clients to this the same way?

No, I just considered it as a bug

eleriaqueen:
I know zero C-Sharp but this code may be related to topic:
[UnityRO - NpcBoxMenuController.cs](https://github.com/guilhermelhr/unityro/blob/dc741a545304265a90d8c27b40a903c6ab316f74/UnityClient/Assets/Scenes/Map/NpcBoxMenuController.cs#L10)

vE5li:
No longer needed since https://github.com/vE5li/korangar/commit/11bc40d4978e264875e36ad068a75842c361c512

PR #28, opened 2023-07-22
Fix stop_measurement panic

gustavobca:
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src\debug\profiling\mod.rs:128:72


vE5li:
@gustavobca Thank you for all the code you are contributing but please pay some attention to formatting (you can run `cargo fmt --all` to properly format before committing) and consistency (style of log messages for example).

Issue #39, opened 2023-07-25
With left-mousebutton kept down the player character should keep walking


eleriaqueen:
Currently keeping left-click down does nothing more than the initial move, character stops at initial destination without changing its course.
On official client keeping left-click down makes the character move until an obstacle is in the way, or until the user lets go of the button.

Short gif of me clicking once and not letting go (official client):
![keep-walking](https://github.com/vE5li/korangar/assets/4411178/e21f90ad-4b48-4ef7-ac44-7eb1cb34bfd3)

vE5li:
Fixed as of #40

PR #40, opened 2023-07-26
Keep walking while LMB is held down

eleriaqueen:
This is a proposed solution for #39

In this gif I keep Left Mouse Button pressed character keeps moving. No accidentally interacting with NPC while button is pressed either.

keep-walking-solution

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -363,7 +363,7 @@ impl InputSystem {
             }
             MouseInputMode::ClickInterface => interface.set_mouse_cursor_state(MouseCursorState::Click, client_tick),
             MouseInputMode::None => {}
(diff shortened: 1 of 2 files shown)


vE5li:
Looks fundamentally good, but you should store the last destination and check if the location you want to move to is the same as the one that's stored. Because the way the code is right now you will request to move every single frame

eleriaqueen:
Hmm I don't know how to do what you said, I guess I would need to declare either two static variables (last_destination_x/y) or define a struct LastDestination and declare it as static but then is input/mod.rs the right place for it ? . Maybe you could do it since at the moment for me coding in Rust is like me trying to swim while flailing my arms in all directions.

vE5li:
No problem, I can implement the checks

PR #46, opened 2023-07-30
Load ron file to load game files with priority

aleDsz:


src/loaders/config/mod.rs
```diff
@@ -0,0 +1,55 @@
...
+                print_debug!(format!(
+                    "reading from order {MAGENTA}{order}{NONE} the game file from path {CYAN}{path}{NONE}"
+                ));
+
+                let index = order.parse::<usize>().unwrap();
+                game_files_list.insert(index, path.trim().to_string());
+            }
+        } else {
+            game_files_list_file.expect(&format!("Failed to read {GAME_FILES_LIST_PATH}"));
+        }
```
vE5li, line 47: This can actually be expressed very nicely in Rust using a `let-else` statement. The code would then look like
```rust
        let Ok(ini_file) = game_files_list_file else {
            panic!("failed to read {MAGENTA}{GAME_FILES_LIST_PATH}{NONE}");
        };

        // continue working with `ini_file`
        // ..
```
The advantage here is that you don't need to unwrap or check for `is_ok()`, but can just model the happy path.

Note that I usually start panic messages with a lowercase letter

src/loaders/config/mod.rs
```diff
@@ -0,0 +1,55 @@
...
+            timer.stop();
+
+            for (order, path) in data_section.iter() {
+                #[cfg(feature = "debug")]
+                print_debug!(format!(
+                    "reading from order {MAGENTA}{order}{NONE} the game file from path {CYAN}{path}{NONE}"
+                ));
+
+                let index = order.parse::<usize>().unwrap();
+                game_files_list.insert(index, path.trim().to_string());
```
vE5li, line 43: I am a bit scared of this line because I think that the official files can also skip certain numbers. If we did that here the code would panic.

src/main.rs
```diff
@@ -189,9 +198,9 @@ fn main() {
 
     let mut game_file_loader = GameFileLoader::default();
 
-    game_file_loader.add_archive("data.grf".to_string());
-    game_file_loader.add_archive("rdata.grf".to_string());
-    game_file_loader.add_archive("korangar.grf".to_string());
+    for game_file_name in config_loader.game_files_list {
+        game_file_loader.add_archive(game_file_name.to_string());
+    }
```
vE5li, line 203: I like the idea of hiding this even more, so maybe doing something like
```rust
config_loader.add_archives(&mut game_file_loader);
```
That would also allow you to make the `game_file_list` private, which is generally desireable

vE5li (commented):
I added some comments but overall I am very please with this PR :+1: 

I was thinking about what the actual advantage is to using in `INI` file here, because it now occurs to me it is not really well suited for this sort of application. The fact that we have to define key-value pairs when what we really want is just a list bothers me a bit.

So I have an alternative suggestion: Let's use [RON](https://github.com/ron-rs/ron).
This would allow us to write the file like this
```rust
archives: ["korangar.grf", "rdata.grf", "data.grf"]
```
So now we don't need to define arbitrary keys any more and also you could remove the whole `game_files_list.reverse();` operation, because you can just iterate over the array backwards.

Another big benefit is that RON uses [serde](https://github.com/serde-rs/serde) as a backend, so we could *super* easily change the file type in the future without changing any code.

I personally really like this approach and I would really love to hear what you think about it :)

src/loaders/gamefile/mod.rs
```diff
@@ -193,13 +196,40 @@ impl GameArchive {
...
+
 impl GameFileLoader {
+    pub fn add_archives_from_settings(&mut self) {
+        #[cfg(feature = "debug")]
+        let timer = Timer::new_dynamic(format!("load game files list from settings"));
+
+        for path in &self.game_file_settings.archives {
+            let game_archive = GameArchive::load(&path, &mut self.lua_files);
+            self.archives.insert(0, game_archive);
+        }
```
aleDsz, line 227: I'm not so happy with this implementation, do you have suggestions? Borrowing the mutable to use `Self.add_archive(self, path)` doesn't work and, I dunno how to implement it in a better way.

vE5li: I also ran into this issue when looking into your PR. I ended up giving the `add_archive` function the actual `GameArchive` object instead of a borrowed `&str` to fix it

PR #50, opened 2023-08-04
Add baby jobs to sprite lookup table

eleriaqueen:
I tested them all using the job change command, even though they use the same sprites as their adult counterpart.
Official ragnarok probably scales down these sprites to make them look like babies.


eleriaqueen:
Woops I forgot that 3rd classes and Expanded classes can be babies too!

vE5li:
Don't worry about the tests failing btw, it's a Rust issue

eleriaqueen:
I left out baby Summoner because Korangar doesn't seem to support Doram race at the moment.

PR #56, opened 2023-08-26
New Archive API

ark487168:
The main issue addressed by this PR is memory usage in Korangar. Currently, all GRF archives are read from disk and cached in memory. While convenient, this approach gives a high usage of RAM.

GameFileLoader module has been refactored, getting out all code related to GRF code. Instead GameFileLoader now works with type implementing the `Archive` trait.

Any type implementing the `Archive` trait manages a storage for individual game assets. It's been added two different types:
- `GrfFileArchive` for accesing assets inside GRF files.
- `FolderArchive` for accessing assets inside OS folders.

`GrfFileArchive` now stores the OS file handler of the GRF file, and each time an asset is requested, proceed to retrieve from disk. Load times has increased lightly, but the usage of memory has been reduced by approximately 75%. Implementation of cache has been removed, but should be trivial to reimplement if needed.

Also, the `settings` module has been added. Using [clap](https://docs.rs/clap/latest/clap/), korangar can use command line arguments to specify some custom behaviour. Some of the commands added has been:

- `game-archive-list` - Determines the location of the config file indication GRF files. (Default: `client/game_archives.ron`)
- `patched-lua-files-location` - Client unify the bytecode of some of the lua files. This option indicates the location of the patched files (Default: lua_files.grf).

While using `patched-lua-files-location` it's possible to use a folder location (like `client/luafiles/`) so all files will be written to a OS folder instead of a GRF file. Should be useful for debug propurses.


vE5li:
Thank you @Duckwhale  for your comments :heart:
As for the `mixcrypt` thing, there is already an issue for it, so I will just post the link over there. This PR is already quite big so I would like to avoid tackling another topic.

PR #58, opened 2023-08-28
Fixed #57 caused by overflow in length_hint while reading sprite data

### LINKED ISSUE #57: Client crash while loading sprites with version 2.1
Since version 2.1, sprites using indexed-color bitmaps stores the BMP segment using Run-length encoding.

Sprite loader module does not have this in consideration, causing a client crash anytime one of those sprites are loaded.

ark487168:
While reading some sprite data using `ByteConvertable` proc_macro, a helper attribute is used to give hints about the length of the buffer containing data of `RgbaImageData`. This hint was being calculated multiplying `width` and `height` attributes of said struct, both declared as u16, and causing a overflow while calculating the total size of the buffer.

In order to calculate the size of the buffer, both parameters are casted to u32 before the multiplication.

### DIFF
--- a/src/loaders/map/data.rs
+++ b/src/loaders/map/data.rs
@@ -77,7 +77,7 @@ pub struct GroundData {
     pub surface_count: i32,
     #[repeating(self.surface_count)]
     pub surfaces: Vec<Surface>,
-    #[repeating(self.width * self.height)]
+    #[repeating(self.width as usize * self.height as usize)]
     pub ground_tiles: Vec<GroundTile>,
 }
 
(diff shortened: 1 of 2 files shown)


vE5li:
Good catch, is this related to #57?

vE5li:
As for the code: I would prefer casting to a `usize` since that is what is happening behind the scenes as well.

Also there are in total 4 occurrences of `#[repeating(self.width * self.height)]` in the code base, three in the sprite loader and one in the map loader. It would be awesome if you could fix the other ones as well

ark487168:
> Good catch, is this related to #57?

I think so. Bug appeared on a particular sprite that, coincidentally, had version 2.1. This got me thinking the bug was about RLE encoding. However, RLE decoding happens while reading `EncodedData` from `PaletteImageData`.

Anyhow, once this patch has been applied, the conflicting sprite seems to work okay-ish, despite presenting some visual glitches in some frames. But this PR fixes the crashes thus improving stability.

PR #61, opened 2023-09-07
Implements reading and parsing `clientinfo` files

aleDsz:
It also adds a new window:

### Select Service window

image


src/loaders/xml/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+impl XMLFileLoader {
+    pub fn load_client_info(&self, game_file_loader: &mut GameFileLoader) -> ClientInfo {
+        let clientinfo = game_file_loader.get("data\\clientinfo.xml");
+        let sclientinfo = game_file_loader.get("data\\sclientinfo.xml");
+
+        let src = match (sclientinfo, clientinfo) {
+            (Ok(data), _) => String::from_utf8(data).unwrap(),
+            (_, Ok(data)) => String::from_utf8(data).unwrap(),
+            _ => panic!("FOO"),
+        };
```
aleDsz, line 24: I remember @vE5li mentioned to read one before reading another one (if it doesn't exist). In this case, what would you suggest to?

src/loaders/xml/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+
+        // TODO: Make it work with euc-kr, since it panics
+        // with error: "Unsupported encoding: euc-kr"
+        let replaced_src = src.replace("euc-kr", "utf8");
```
aleDsz, line 28: I tried to use the [`ParserConfig2` from `xml-rs`](https://github.com/netvl/xml-rs/blob/16348a470ac9e9903606925a7bf271c0627261e9/src/reader/config.rs#L198-L199) to override the file encoding when reading the `clientinfo.xml`, but it still fails with every available option.

I would keep it as-is until we find a solution for this.

src/loaders/mod.rs
```diff
@@ -28,3 +29,4 @@ pub use self::sprite::*;
 pub use self::stream::ByteStream;
 pub use self::texture::TextureLoader;
 pub use self::version::{InternalVersion, MajorFirst, MinorFirst, Version};
+pub use self::xml::*;
```
vE5li, line 32: Give that there are only `Connection` and `XMLFileLoader` here to export, I would prefer if we remove the wildcard export.

src/loaders/xml/client_info.rs
```diff
@@ -0,0 +1,282 @@
...
+}
+
+fn bool_deserializer<'de, D>(d: D) -> Result<bool, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let s = String::deserialize(d)?;
+    match &s[..] {
+        "" => Ok(true),
+        _ => panic!("don't do that"),
```
vE5li, line 280: Something like `"boolean tags may not have any data inside"` seems more fitting xD

src/loaders/xml/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+
+        let config = ParserConfig::new().trim_whitespace(true);
+        let event_reader = EventReader::new_with_config(replaced_src.as_bytes(), config);
+        let client_info = ClientInfo::deserialize(&mut Deserializer::new(event_reader)).unwrap();
+
+        #[cfg(feature = "debug")]
+        print_debug!("reading client info:\n{}{:#?}{}", MAGENTA, client_info, NONE);
```
vE5li, line 35: I would like to remove this as it produces too many lines in the logs and is generally a bit inconsistent with the rest of the logging. My suggestion would be to add a
```rust
#[cfg(feature = "debug")]
let timer = Timer::new("read clientinfo");
```
in the beginning and
```rust
#[cfg(feature = "debug")]
timer.stop();
```
at the end.

src/loaders/xml/mod.rs
vE5li: This can be very nicely expressed with some functional style Rust
```rust
let clientinfo = game_file_loader
    .get("data\\sclientinfo.xml")
    .or_else(|_| game_file_loader.get("data\\clientinfo.xml"))
    .expect("failed to find clientinfo");
```
In this case, we only load `clientinfo.xml` if `sclsclientinfo.xml` was not found, thereby saving some computation. And since all of this works on `Result`s we can easily `expect` that at least one of them is present.

src/loaders/xml/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+
+        client_info
+    }
+}
```
vE5li, line 39: It doesn't really make a lot of sense that `XMLFileLoader` is a struct, in the sense that we only ever construct it to use a single method once (also it doesn't use the `&self` parameter). It would be nicer if we could just expose the function called `load_client_info` on it's own.

vE5li (commented):
Very nice PR :+1: 
I love how much documentation you added to the clientinfo

src/loaders/xml/client_info.rs
```diff
@@ -0,0 +1,282 @@
...
+        ServerType::Unknown => Err(D::Error::custom("invalid server type {name}")),
+        value => Ok(value),
+    }
+}
+
+fn bool_deserializer<'de, D>(d: D) -> Result<bool, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let s = String::deserialize(d)?;
```
vE5li, line 277: Please use a more descriptive identifier like `value` or `data` or something.

src/loaders/xml/mod.rs
aleDsz: Makes sense, tbh.

vE5li (approved):
Amazing work :+1:

PR #62, opened 2023-09-29
Show Select Server window after successfully login

aleDsz:
https://github.com/vE5li/korangar/assets/6402997/4e97be7f-0ca8-48aa-99d6-b98620747015

### DIFF
--- a/src/network/mod.rs
+++ b/src/network/mod.rs
@@ -466,7 +466,7 @@ impl FromBytes for Ipv4Addr {
 }
 
 #[derive(Clone, Debug, Named, FromBytes, FixedByteSize, PrototypeElement)]
-struct CharacterServerInformation {
+pub struct CharacterServerInformation {
     pub server_ip: Ipv4Addr,
     pub server_port: u16,
     pub server_name: [u8; 20],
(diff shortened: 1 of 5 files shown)


src/network/mod.rs
```diff
@@ -2861,42 +2863,71 @@ impl NetworkingSystem {
...
-            .character_server_information
-            .into_iter()
-            .next()
-            .ok_or("no character server available")?;
+        let character_server_information_list: Vec<CharacterServerInformation> =
+            login_server_login_success_packet.character_server_information.into();
+
+        if character_server_information_list.len() == 0 {
+            return Err("no character server available".to_string());
+        }
```
aleDsz, line 2876: I'm not so sure about this part, but mostly of that are working great.

vE5li: I updated the code slightly to clean up the `into` thing a bit but I think returning an `Err` here is actually perfect.

Side note: I would really recommend you to install `clippy`

vE5li (approved):
Thank you @aleDsz, really good code! :+1:

Edit: And such a clean commit history :heart_eyes:

PR #69, opened 2024-02-29
change internal representation of Color to f32

### LINKED ISSUE #67: Store `Color` as `f32` internally
Currently [`Color`](https://github.com/vE5li/korangar/blob/e77cd0311127d8167584950762da2de86ce4e769/src/graphics/color.rs#L5) is stored as a `u8` for each color channel, but is most often used as an `f32`.

To decrease the amount of processing needed it would be nice to store the color channels as `f32`.

TheOptik:
Closes #67 

Sorry for the old PR, i accidentally created it from main

### DIFF
--- a/src/interface/windows/mod.rs
+++ b/src/interface/windows/mod.rs
@@ -51,7 +51,8 @@ impl Window {
     }
 
     pub fn has_transparency(&self, theme: &InterfaceTheme) -> bool {
-        self.get_background_color(theme).alpha_as_u8() != 255
+        const TRANSPARENCY_THRESHOLD: f32 = 0.999;
+        self.get_background_color(theme).alpha < TRANSPARENCY_THRESHOLD
     }
 
     pub fn is_closable(&self) -> bool {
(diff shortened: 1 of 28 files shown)


src/interface/windows/mod.rs
```diff
@@ -51,7 +51,7 @@ impl Window {
     }
 
     pub fn has_transparency(&self, theme: &InterfaceTheme) -> bool {
-        self.get_background_color(theme).alpha != 255
+        self.get_background_color(theme).alpha_as_u8() != 255
```
vE5li, line 54: I think it would make sense here to compare against `1.0` so we don't bring floating point inaccuracy into the calculation

vE5li (changes requested):
One minor comment, apart from that it looks excellent :)

src/interface/windows/mod.rs
TheOptik: Wouldn't it make more sense to compare using a small epsilon?
You are totally right that this isn't a good check, but i am not confident that comparing to 1.0 is a better approach.
I'll make a change and you just tell me if you are happy with it or not :)

vE5li: Very true!
I like the solution a lot :+1:

vE5li (approved):

PR #83, opened 2024-06-18
Adds nix flake devshell

arpd:
I've only tested this on aarch64 macos, but
should also work on any linux environment.
May have to add `LD_LIBRARY_PATH` in that case.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -16,9 +16,7 @@
         devShell = pkgs.mkShell
           {
             nativeBuildInputs = with pkgs; [
-              (rust-bin.selectLatestNightlyWith (toolchain: toolchain.default.override {
-                extensions = [ "rust-src" "rust-analyzer" "miri" ];
-              }))
+              (rust-bin.fromRustupToolchainFile ./rust-toolchain.toml)
               pkg-config
             ];
             buildInputs = with pkgs;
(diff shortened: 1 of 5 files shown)


flake.nix
```diff
@@ -0,0 +1,63 @@
...
+      let
+        overlays = [ (import rust-overlay) ];
+        pkgs = (import nixpkgs) { inherit system overlays; };
+      in
+      {
+        # Provide a shell with all dependencies needed to build `koragnar::*`
+        devShell = pkgs.mkShell
+          {
+            nativeBuildInputs = with pkgs; [
+              (rust-bin.selectLatestNightlyWith (toolchain: toolchain.default.override {
```
vE5li, line 19: I was trying to find some refenrence on `selectLatestNightlyWith` to figure out if it respects the `rust-toolchain.toml`, but I believe it does not, correct?

vE5li (commented):
Okay first of all, very cool! I've been meaning to add a `flake.nix` for a while but I just didn't know how to do it right. I will probably ask some more questions to fully understand this before merging.

First one right away, does it make a difference if we move these files to the `korangar` directory?
I'm trying to keep the client specific stuff a bit separate from the generic RO stuff, so I would prefer that if possible.

arpd:
I've updated the PR as per your suggestions @vE5li
You can use it either with `direnv`, or `nix develop -c $SHELL` and successfully build and run the project.
It uses the `rust-toolchain.toml` version of the toolchain.

vE5li:
Very nice work! :)

PR #84, opened 2024-08-20
Update nightly toolchain

hasenbanck:
- Updates the nightly toolchain to the most recent one.
- Uses the new prelude functions that came with 1.80.
- Use LazyLock instead of lazy_static.
- Remove unused features.
- Update used features that changed.
- Removed unneeded "allow(suspicious_auto_trait_impls)".
- Synced the order of attributes between all crates.


vE5li:
Thank you!
There seems to be a regression in the newer version of `rustfmt`, but it's not a problem :+1:

PR #87, opened 2024-08-23
Improve the profiler.

hasenbanck:
- No use of unsafe and possible UB.
 - Allocations are minimized.
 - All measurements of a frame are saved continuously.
 - Faster length calculation and index operation on ring buffer.
 - Added tests for the container.
 
Benchmark:

```
start_frame() with drop:

old = 112.10 ns
new = 60.176 ns

start_measurement() with drop:

old = 93.523 ns
new = 96.892 ns
```

start_frame() is a lot faster, start_measurement() is a little bit slower, but overall the consistent performance should improve, since once all frame slots are filled, we won't hit the allocator anymore. Packets still hit the allocator, because the boxed trait object can't be re-used.

### DIFF
--- a/korangar/src/interface/elements/profiler/inspector.rs
+++ b/korangar/src/interface/elements/profiler/inspector.rs
@@ -1,7 +1,7 @@
 use std::time::{Duration, Instant};
 
 use derive_new::new;
-use korangar_debug::profiling::Measurement;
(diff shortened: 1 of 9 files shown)


vE5li:
Very nice PR! Super exited to have this on main :)

PR #88, opened 2024-08-23
Remove unsafe code from PacketHistoryCallback.

hasenbanck:
This makes sure that we only initialize the buffer once.

### DIFF
--- a/korangar/src/main.rs
+++ b/korangar/src/main.rs
@@ -331,7 +331,7 @@ fn main() {
         #[cfg(not(feature = "debug"))]
         let mut networking_system = NetworkingSystem::spawn();
         #[cfg(feature = "debug")]
-        let packet_callback = interface::elements::PacketHistoryCallback::new();
+        let packet_callback = interface::elements::PacketHistoryCallback::get_static_instance();
         #[cfg(feature = "debug")]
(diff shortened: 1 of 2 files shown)


korangar/src/main.rs
```diff
@@ -331,11 +331,7 @@ fn main() {
         #[cfg(not(feature = "debug"))]
         let mut networking_system = NetworkingSystem::spawn();
         #[cfg(feature = "debug")]
-        let packet_callback = {
-            // SAFETY: This function leaks memory, but it's fine since we only call
-            // it once.
-            unsafe { interface::elements::PacketHistoryCallback::new() }
-        };
+        let packet_callback = interface::elements::PacketHistoryCallback::new();
```
vE5li, line 334: I don't love the fact that to the caller it looks like you are getting a new, unaliased `PacketHistoryCallback`. I was thinking of renaming the new function to something like `static_instance` or `get_singleton`. What do you think?

hasenbanck: Fine by me! How about `new_from_static_instace`?

vE5li: Hm, `get_static_instance`?

hasenbanck: Also fine!

PR #89, opened 2024-08-23
Improve the color parsing.

hasenbanck:
Remove the need for unsafe code, the intermediate buffered bytes and also removed the cloning of the Chars iterator.

Instead we do some tricks with slices. We re-create the Chars iterator in the case, that we successfully parse a color. There is the unstable "iter_advance_by", but after looking at it's code I think re-creating the Chars iterator is more efficient (it's only two pointer that we set).

### DIFF
--- a/korangar/src/loaders/font/mod.rs
+++ b/korangar/src/loaders/font/mod.rs
@@ -115,6 +115,7 @@ fn layout_paragraph(font: &Font<'static>, scale: Scale, width: f32, text: &str, 
     let mut last_glyph_id = None;
     let mut color = default_color;
     let mut chars = text.chars();
+    let text_address = text.as_ptr() as usize;
 
     while let Some(character) = chars.next() {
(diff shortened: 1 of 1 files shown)


hasenbanck:
Ah damn, don't merge this PR and close it for now. We are too trustworthy for the incoming data. We there are uneven byte characters, the function will blow up in out face:

[Example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ed3bb5a8d0708d63a5e02de59e838e98)

hasenbanck:
Okay, this version is a hybrid of both approaches. I used the old and correct way to see, if the next 6 characters are 1 byte hexdigits. If so, we then do slice manipulation, to select the hexdigits and set the chars iterator to the remaining text.

Cloning and creating the chars iterator should be cheap, since it's just copying two pointers each time.

PR #90, opened 2024-08-24
Prune and update dependencies and improve maintainability.

hasenbanck:
- Track all versions inside the workspace Cargo.toml
- Updated dependencies that had no breaking changed
- Removed default features for image and tokio dependency
- Selectively enabled used features for image and tokio
- Removed unused crates inside the main project
- Moved dependencies that examples use behind a example feature flag
- Updated documentation for example feature flag
- Updated example to only use tokio features that we also use in the crate itself

This PR should ease maintenance and reduces build times.

```
Release 
Before: 259 dependencies (0m56,504s)
After:  241 dependencies (0m52,895s)

Develop
Before: 230 dependencies (0m46,335s)
After:  215 dependencies (0m45,612s)
```

### DIFF
--- a/korangar_networking/README.md
+++ b/korangar_networking/README.md
@@ -13,7 +13,7 @@ It connects to the server as a client and uses `Ollama` to generate responses to
 To run this example you need to set `USERNAME`, `PASSWORD`, and `CHARACTER_NAME` in the source code (you might also have to adjust `OLLAMA_ENDPOINT` and `OLLAMA_MODEL` depending on your setup).
 Afterwards you can run it with
 ```bash
-cargo run --example ollama-chat-bot
+cargo run --example ollama-chat-bot --features=example
 ```
 
 ##### Note: Make sure that Ollama is serving and the model specified in `OLLAMA_MODEL` is installed, otherwise you will get a `404`
--- a/korangar_networking/README.md
+++ b/korangar_networking/README.md
@@ -13,7 +13,7 @@ It connects to the server as a client and uses `Ollama` to generate responses to
 To run this example you need to set `USERNAME`, `PASSWORD`, and `CHARACTER_NAME` in the source code (you might also have to adjust `OLLAMA_ENDPOINT` and `OLLAMA_MODEL` depending on your setup).
 Afterwards you can run it with
 ```bash
-cargo run --example ollama-chat-bot --features=example
+cargo run --example ollama-chat-bot
 ```
 
 ##### Note: Make sure that Ollama is serving and the model specified in `OLLAMA_MODEL` is installed, otherwise you will get a `404`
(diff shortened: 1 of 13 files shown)


korangar_networking/examples/ollama-chat-bot.rs
```diff
@@ -59,7 +59,7 @@ impl MessageHistory {
     }
 }
 
-#[tokio::main]
+#[tokio::main(flavor = "current_thread")]
```
vE5li, line 62: What is the reason for this change?

korangar_networking/README.md
```diff
@@ -13,7 +13,7 @@ It connects to the server as a client and uses `Ollama` to generate responses to
 To run this example you need to set `USERNAME`, `PASSWORD`, and `CHARACTER_NAME` in the source code (you might also have to adjust `OLLAMA_ENDPOINT` and `OLLAMA_MODEL` depending on your setup).
 Afterwards you can run it with
 ```bash
-cargo run --example ollama-chat-bot
+cargo run --example ollama-chat-bot --features=example
```
vE5li, line 16: Can you maybe explain what the advantage of this feature is over having the `dev-dependencies` as non optional?

I think regardless I would prefer not to require any features when running the examples.

vE5li (commented):
Looks good, just two small questions

korangar_networking/examples/ollama-chat-bot.rs
hasenbanck: We only use the single-threaded (ST) runtime in the project, so to not include the multi-threaded (MT) runtime also, we also use the ST variant here. But if you don't want to use the feature as state below, this can be returned to use the MT runtime.

korangar_networking/README.md
hasenbanck: My main goal was to reduce build times when running tests. When running all tests inside the workspace and not having this feature, we would need to include the dependencies that are now behind this feature, even though they are only used inside the example and not helpfull in general.

korangar_networking/examples/ollama-chat-bot.rs
vE5li: Okay, I like this :+1:

korangar_networking/README.md
vE5li: While I obviously like reduced build times, for tests I value ease of use over build times. So I would prefer if we removed this feature again and just include the dependencies unconditionally.

hasenbanck: Fine by me! Should I change it or do you want to change it?

vE5li: I will leave it to you :D

hasenbanck: Done!

PR #92, opened 2024-09-17
Implement the basic audio engine.

hasenbanck:
This PR lays the foundation for the audio engine. It implements the main menue sounds, music and ambient, spatial sound. Entity and Effect sound effects will be implemented later. Once the new UI code is finished, I will also start to expose the configuration values (mainly volume control).

The PR contains a "util" library, which will contain code that other libraries might want to use (mainly containers and collision code). I will also use these containers for the render engine later (for async loading etc.). I took the liberty to make the usage of the game archive loader a little bit easier (since we want to use it from async threads later). I also removed the unmaintained collision-rs library and re-use collision code between the frustum culling and the audio engine.

### DIFF
--- a/korangar/src/loaders/gamefile/mod.rs
+++ b/korangar/src/loaders/gamefile/mod.rs
@@ -5,10 +5,11 @@ mod list;
 
 use core::panic;
 use std::path::Path;
-use std::u8;
+use std::sync::RwLock;
 
 #[cfg(feature = "debug")]
 use korangar_debug::logging::{print_debug, Colorize, Timer};
+use korangar_util::{FileLoader, FileNotFoundError};
 
 use self::list::GameArchiveList;
 use super::archive::folder::FolderArchive;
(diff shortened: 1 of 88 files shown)


korangar/src/loaders/gamefile/mod.rs
```diff
@@ -42,6 +39,42 @@ pub struct GameFileLoader {
...
+        let lowercase_path = path.to_lowercase();
+        let result = self
+            .archives
+            .read()
+            .unwrap()
+            .iter()
+            .find_map(|archive| archive.get_file_by_path(&lowercase_path))
+            .ok_or(FileNotFoundError::new(path.to_owned()));
+
+        // TODO: should this be removed in the future or left in for resilience?
```
aleDsz, line 53: We should only crash the application if there's no `.spr`, `.act` or `.gat` files (not sure about `.rsw` and `.gat` ones), since the official client keeps running if there's no image file.

aleDsz: I know this is just a copy-paste from the other file, but just pointing it here.

aleDsz: I had an issue recently with `.wav` files, they only crash if the game sound is enabled

hasenbanck: Can we move this issue out of the scope of this PR?

hasenbanck:
Ok, I'm finished here now.

PR #93, opened 2024-09-23
Add point lights with drop shadows

vE5li:
This PR adds the MVP of drop shadows for point lights. This includes
a fixed amount of point lights that can be rendered with shadows and a
point light manager that takes care of deciding which point light
shadows are most important for the scene.

Additionally it improves the way frustum culling works to play better
with point lights.

Also includes small unrelated cleanups.


korangar/src/main.rs
```diff
@@ -87,6 +90,9 @@ fn main() {
     const DEFAULT_BACKGROUND_MUSIC: Option<&str> = Some("background_music\\01.mp3");
     const MAIN_MENU_CLICK_SOUND_EFFECT: &str = "¹öÆ°¼Ò¸®.wav";
 
+    // TODO: This should not be a constant.
```
hasenbanck, line 93: Can you maybe add to the comment what it should be instead or why it's fixed to 6?

korangar/src/main.rs
```diff
@@ -307,9 +314,23 @@ fn main() {
...
-            .map(|_| shadow_renderer.create_render_target(shadow_detail.get().into_resolution()))
-            .collect::<Vec<<ShadowRenderer as Renderer>::Target>>();
+        let mut directional_shadow_targets = (0..max_frame_count)
+            .map(|_| directional_shadow_renderer.create_render_target(shadow_detail.get().into_resolution()))
+            .collect::<Vec<<DirectionalShadowRenderer as Renderer>::Target>>();
+
+        let mut point_shadow_targets = (0..max_frame_count)
+            .map(|_| {
+                (0..NUMBER_OF_POINT_LIGHTS_WITH_SHADOWS)
+                    // TODO: Replace with shadow detail
```
hasenbanck, line 324: Do you still want to do that in this PR?

korangar/src/main.rs
```diff
@@ -879,8 +907,9 @@ fn main() {
                             effect_holder.add_effect(Box::new(EffectWithLight::new(
                                 effect,
                                 frame_timer,
-                                EffectCenter::Entity(entity_id, cgmath::Vector3::new(0.0, 0.0, 0.0)),
+                                EffectCenter::Entity(entity_id, cgmath::Point3::new(0.0, 0.0, 0.0)),
                                 Vector3::new(0.0, 9.0, 0.0),
+                                PointLightId::new(entity_id.0), // FIX: Not unique!!
```
hasenbanck, line 912: Why do we need the uniqueness and how can we create it?

korangar/src/main.rs
```diff
@@ -1610,16 +1661,90 @@ fn main() {
...
+                            // performance.
+
+                            for index in 0..6 {
+                                point_shadow_camera.change_direction(index);
+                                point_shadow_camera.generate_view_projection(Vector2::zero());
+
+                                let (view_matrix, projection_matrix) = point_shadow_camera.view_projection_matrices();
+                                let mut point_shadow_render_pass = point_shadow_target.start(&queue, &mut point_shadow_command_encoder, index, projection_matrix * view_matrix);
+
+                                /* #[cfg_attr(feature = "debug", korangar_debug::debug_condition(render_settings.show_entities))]
```
hasenbanck, line 1702: Do we still need to fix this? Is it related to the TODO above?

korangar/src/main.rs
```diff
@@ -1844,12 +1983,13 @@ fn main() {
...
                 drop(interface_render_pass);
-                drop(shadow_render_pass);
+                drop(directional_shadow_render_pass);
                 drop(geometry_render_pass);
                 drop(screen_render_pass);
                 let picker_command_buffer = picker_target.finish(picker_command_encoder);
                 let interface_command_buffer = interface_target.finish(interface_command_encoder);
-                let shadow_command_buffer = shadow_target.finish(shadow_command_encoder);
+                let directional_shadow_command_buffer = directional_shadow_target.finish(directional_shadow_command_encoder);
+                let point_shadow_command_buffer = point_shadow_target[0].finish(point_shadow_command_encoder); // FIX: Doesn't make any sense with the 0 here
```
hasenbanck, line 1992: Maybe add, that it only works because we don't do anything target specific with the encoder (we only call "finish()" on it).

korangar/src/world/map/mod.rs
```diff
@@ -468,6 +489,7 @@ impl Map {
             MarkerIdentifier::Particle(..) => panic!(),
             // TODO: implement properly
             MarkerIdentifier::Entity(index) => &entities[index],
+            MarkerIdentifier::Shadow(..) => panic!(),
```
hasenbanck, line 492: Maybe we want to use "todo!()" for panics that we want to properly implement?

ragnarok_formats/src/map.rs
```diff
@@ -318,6 +318,12 @@ impl FromBytes for MapResources {
                 ResourceType::LightSource => {
                     let mut light_source = LightSource::from_bytes(byte_stream).trace::<Self>()?;
                     light_source.position.y = -light_source.position.y;
+
+                    // Some light sources have color channels with values bigger than 1.0 (255), so
+                    // we need to clamp them.
+                    // TODO: Does this maybe have a special meaning?
```
hasenbanck, line 324: Yes, if we want to increase the percieved dynamic range and implement an HDR effect in SDR, we need to implement proper tone mapping in our light calculation (for example using Reinhard tone mapping).

korangar/src/graphics/renderers/point_shadow/indicator/indicator.wgsl
```diff
@@ -0,0 +1,87 @@
...
+}
+
+struct FragmentInput {
+    @location(0) world_position: vec4<f32>,
+    @location(1) texture_coordinates: vec2<f32>,
+}
+
+@group(0) @binding(0) var sampled_texture: texture_2d<f32>;
+@group(0) @binding(1) var texture_sampler: sampler;
+@group(1) @binding(0) var<uniform> view_projection: mat4x4<f32>;
```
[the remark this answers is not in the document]
vE5li: Yeah I also thought about doing that, but at least in my head this seemed more performant (?) that using push constants. Just my naive understanding though.

korangar/src/graphics/cameras/point_shadow.rs
```diff
@@ -0,0 +1,85 @@
...
+}
+
+impl PointShadowCamera {
+    const FAR_PLANE: f32 = 256.0;
+    const NEAR_PLANE: f32 = 0.1;
+
+    pub fn new() -> Self {
+        Self {
+            camera_position: Point3::new(0.0, 0.0, 0.0),
+            view_direction: Vector3::new(1.0, 0.0, 0.0),
```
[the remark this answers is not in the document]
vE5li: I like it. I will probably do that outside of this PR because I want to change it in all cameras then.

korangar/src/graphics/renderers/point_shadow/entity/entity.wgsl
```diff
@@ -0,0 +1,121 @@
...
+    let diffuse_color = textureSample(texture, texture_sampler, fragment.texture_coordinates);
+
+    if (diffuse_color.a != 1.0) {
+        discard;
+    }
+
+    let scaled_depth_offset = pow(fragment.depth_offset, 2.0) * constants.depth_offset;
+    // let scaled_curvature_offset = (0.5 - pow(fragment.curvature, 2.0)) * constants.curvature;
+
+    // let linear_z: f32 = nonLinearToLinear(position.z);
```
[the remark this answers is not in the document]
vE5li: This will be used in the future. The entire point shadow entity renderer is actually not used currently.

korangar/src/main.rs
vE5li: I actually didn't mean to but now that you called me out, yes :D

korangar/src/graphics/renderers/point_shadow/geometry/geometry.wgsl
```diff
@@ -0,0 +1,60 @@
+struct Matrices {
+    time: f32,
+}
+
+struct Constants {
+    world: mat4x4<f32>,
+    light_position: vec3<f32>,
```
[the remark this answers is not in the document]
vE5li: Yeah. In my mind it was not an issue because it's at the end of the struct, but you are most likely correct in not trusting the driver.

korangar/src/main.rs
vE5li: Thanks! Done :+1:

ragnarok_formats/src/map.rs
hasenbanck: True, it might be an error in the source files.

PR #97, opened 2024-09-27
Parse Quadtree in MapData 

kokosha:
Implementing Quadtree parser for MapData.
Small fix in some segment fault.

### DIFF
--- a/ragnarok_formats/src/map.rs
+++ b/ragnarok_formats/src/map.rs
@@ -1,4 +1,7 @@
+use std::collections::VecDeque;
+
 use cgmath::{Point3, Vector3};
+use korangar_interface::elements::PrototypeElement;
 use ragnarok_bytes::{ByteConvertable, ByteStream, ConversionError, ConversionResult, ConversionResultExt, FromBytes, ToBytes};
 
 use crate::color::{ColorBGRA, ColorRGB};
@@ -40,10 +43,19 @@ pub struct MapData {
     pub ground_left: Option<i32>,
     #[version_equals_or_above(1, 6)]
     pub ground_right: Option<i32>,
-    // TODO: verify version
-    //`#[version_equals_or_above(2, 6)]
-    //pub quad_tree: QuadTree,
+    // TODO: Parse remaining fields
     pub resources: MapResources,
+    #[version_equals_or_above(2, 1)]
+    pub quadtree: Option<QuadTreeData>,
+}
+
+#[derive(Clone, PrototypeElement)]
+pub struct QuadTreeData {
+    pub max: [f32; 3],
+    pub min: [f32; 3],
+    pub half_size: [f32; 3],
+    pub center: [f32; 3],
+    pub children: Vec<QuadTreeData>,
 }
 
 bitflags::bitflags! {
@@ -53,6 +65,72 @@ bitflags::bitflags! {
         const WATER = 0b00000010;
         const SNIPABLE = 0b00000100;
         const CLIFF = 0b00001000;
+    }
+}
+
+impl FromBytes for QuadTreeData {
+    fn from_bytes<Meta>(byte_stream: &mut ByteStream<Meta>) -> ConversionResult<Self> {
+        const MAX_DEPTH: usize = 5;
+        const CHILD_COUNT: usize = 4;
+
+        // Helper struct to keep node accesses readable.
+        struct Node {
+            data: QuadTreeData,
+            parent_index: usize,
+        }
+
+        // Simulate a DFS using a stack.
+        let mut stack = VecDeque::from([(0, 0)]);
+        let mut nodes = Vec::new();
+
+        while let Some((parent_index, depth)) = stack.pop_back() {
+            let max = FromBytes::from_bytes(byte_stream).trace::<Self>()?;
+            let min = FromBytes::from_bytes(byte_stream).trace::<Self>()?;
+            let half_size = FromBytes::from_bytes(byte_stream).trace::<Self>()?;
+            let center = FromBytes::from_bytes(byte_stream).trace::<Self>()?;
+
+            let children = match depth < MAX_DEPTH {
+                true => Vec::with_capacity(CHILD_COUNT),
+                false => Vec::new(),
+            };
+
+            // Add next node lookups to the stack.
+            if depth < MAX_DEPTH {
+                let self_index = nodes.len();
+
+                for _counter in 0..CHILD_COUNT {
+                    stack.push_back((self_index, depth + 1));
+                }
+            }
+
+            let data = QuadTreeData {
+                max,
+                min,
+                half_size,
+                center,
+                children,
+            };
+            nodes.push(Node { data, parent_index });
+        }
+
+        // Go back to front and assign nodes to their parents. Node 0 is the root
+        // node so we skip that.
+        while nodes.len() > 1 {
+            let node = nodes.pop().unwrap();
+
+            let children = &mut nodes[node.parent_index].data.children;
+            children.push(node.data);
+
+            // We reverse the nodes once all of them are collected rather than inserting
+            // them at the start of the vector to avoid unnecessary shifting of
+            // other items. This is purely an optimization.
+            if children.len() == 4 {
+                children.reverse();
+            }
+        }
+
+        let root_node = nodes.pop().unwrap().data;
+        Ok(root_node)
     }
 }
 


ragnarok_formats/src/map.rs
```diff
@@ -56,6 +74,54 @@ bitflags::bitflags! {
...
+        // go from back to start to set the quadtree children in correct order
+        for index in (1..child_index as usize).rev() {
+            let quadtree = quadtree_vec.pop().unwrap();
+            quadtree_vec[quadtree_parent[index] as usize].child.push(quadtree);
+            if quadtree_vec[quadtree_parent[index] as usize].child.len() == 4 {
+                quadtree_vec[quadtree_parent[index] as usize].child.reverse();
+            }
+        }
+        quadtree_vec[0].child.reverse();
+        Ok(quadtree_vec.first().unwrap().clone())
```
vE5li, line 120: I was having some trouble understanding what is happening here, so I reworked this function a bit:

```rust
const MAX_DEPTH: usize = 5;
const CHILD_COUNT: usize = 4;

// Simulate a DFS using a stack.
let mut stack = VecDeque::<(usize, usize)>::from([(0, 0)]);
let mut nodes: Vec<QuadTreeData> = Vec::new();
let mut node_parents: Vec<usize> = Vec::new();

while let Some((parent_node, depth)) = stack.pop_front() {
    let max = FromBytes::from_bytes(byte_stream).trace::<Self>()?;
    let min = FromBytes::from_bytes(byte_stream).trace::<Self>()?;
    let half_size = FromBytes::from_bytes(byte_stream).trace::<Self>()?;
    let center = FromBytes::from_bytes(byte_stream).trace::<Self>()?;
    let children: Vec<QuadTreeData> = match depth < MAX_DEPTH {
        true => Vec::with_capacity(CHILD_COUNT),
        false => Vec::new(),
    };

    nodes.push(QuadTreeData {
        max,
        min,
        half_size,
        center,
        children,
    });

    if depth < MAX_DEPTH {
        for _counter in 0..CHILD_COUNT {
            node_parents.push(parent_node);
            let self_index = node_parents.len();
            stack.push_back((self_index, depth + 1));
        }
    }
}

// Iterate nodes backwards and assign them the their parents.
for parent_index in node_parents.into_iter().rev() {
    let child = nodes.pop().unwrap();
    let parent = &mut nodes[parent_index];

    parent.children.insert(0, child);
}

assert_eq!(nodes.len(), 1);
nodes.pop().unwrap()
```

Notable changes are:
- Adding constants for magic values (`MAX_DEPTH`, `CHILD_COUNT`)
- Renaming some variables to better express what they do
- Using `FromBytes::from_bytes` to let the Rust type system figure out the exact type
- Allocating vectors with the correct size to avoid reallocation
- Removed `child_index` since it can always be derived from the length of the `quadtree_parent`/`node_parents`
- Use `insert(0, ..)` for the children instead of `push(..)` followed by a `reverse`
- Not cloning the root node since we can just remove it from the vector

#### Questions

I also believe there were two errors in the logic of the original function.

1: You skip index `0` in line `112` but you also never iterate over `child_index`, only `child_index - 1` (because of the non-inclusive `..`) which left 2 nodes in the `quadtree_vec` at the end of the loop. If I understand correctly, there should only be 1 left (the root node).

2: The call to `quadtree_vec[0].child.reverse();` in line `119` reverses the ordering of child nodes a second time, since the first `reverse` is already happening in the loop above. I assume the root node should have the same ordering as the other nodes, so this line is incorrect I believe. Please correct me if I'm wrong!

vE5li (commented):
First of all, thank you for your contribution!

I left you some comments, regarding your initial changes. I would request you open a separate PR for the death/reviving changes, as that will make it much easier to merge the changes.

Small hint, you should run `cargo fmt --all` before committing, otherwise there is a very high chance the pipeline will fail.

korangar_interface/src/layout/bound.rs
```diff
@@ -172,7 +172,7 @@ impl SizeBound {
     {
         let half_size = size.halved();
         let left = f32::clamp(position.left(), -half_size.width(), available.width() - half_size.width());
-        let top = f32::clamp(position.top(), 0.0, available.height() - 30.0);
+        let top = f32::clamp(position.top(), 0.0, (available.height() - 30.0).max(0.0));
```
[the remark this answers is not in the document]
kokosha: The game crashes when the available.height() is 0, because in the clamp function the min is 0.0, and the max is -30.0, that trigger the panic and crashes the game.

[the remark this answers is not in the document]
vE5li: Okay, this `-30.0` seems to be completely wrong anyway. Feel free to remove that :+1:

ragnarok_formats/src/map.rs
kokosha: The first part of the algorithm is trying to construct the nodes from the QuadTree by generating recursively the nodes by a DFS using an stack. 

The second part of algorithm after the DFS, we construct the child relationship of the nodes from QuadTree by using the information about the parent that are gathered through the DFS. The way that I construct the child relationship is by running in the reverse order of DFS transversal order, you can do this by just following in the reverse order of how the nodes are being allocated in `quadtree_vec`, and assign the children to the correct parent.

1: For the Question 1, I forgot to iterate in `child_index` and should have used range equal (..=). Need to fix it.

2: For the Question 2, the second one I think I am correct, but I was using `reverse` whenever find out that the parent has four child, this means the child is well constructed and we can apply the `reverse`. And this condition was not being check in the root node, so I need to apply particularly to this case.

I need to create a test to check if the code is correct, maybe I am slipping in more stuff there.

ragnarok_formats/src/map.rs
```diff
@@ -41,9 +45,23 @@ pub struct MapData {
     #[version_equals_or_above(1, 6)]
     pub ground_right: Option<i32>,
     // TODO: verify version
-    //`#[version_equals_or_above(2, 6)]
-    //pub quad_tree: QuadTree,
+    //#[version_equals_or_above(2, 7)]
+    //#[new_derive]
+    //pub something_count: Option<u32>,
+    //#[repeating(something_count)]
+    //pub something: Vec<i32>,
```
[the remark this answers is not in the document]
kokosha: There is more stuff to implement, I am going to maintain the comment, so there is a warning that there is a future implementation needed.

ragnarok_formats/src/map.rs
vE5li: > And this condition was not being check in the root node

Why would it not be checked? The root node will also have four children after the loop right? Unless it didn't because of the first error, which is very possible.

kokosha: In the way that you have implemented the second part, the check is not needed anymore.

Oh, and I also think I am doing stack wrong should be `stack.push_front((self_index, depth + 1));` instead of `stack.push_back((self_index, depth + 1));`. I have an C++ bad muscle memory of typing because of vectors, .push_back()...
 I am going to push your modification and I would try to create a test, but it is not easy to create a test for this and to check if is everything all right. 
I will create another branch for the death/ reviving part.

vE5li: Very nice, thanks :)

kokosha (commented):
Tested the QuadTree.
It is working.

ragnarok_formats/src/map.rs
```diff
@@ -55,6 +72,79 @@ bitflags::bitflags! {
...
+            nodes.insert(parent_index, parent);
+        }
+
+        // The root haven't reverse yet, so we need to reverse it
+        // Observe that the root isn't in parent_index, only the childs of root.
+        assert_eq!(nodes.len(), 1);
+        let mut root: QuadTreeData = nodes.remove(&0).unwrap();
+        root.children.reverse();
+
+        Ok(root)
```
vE5li, line 145: I did some refactoring on the newer version again, please check it out:

```rust
        // Helper struct to keep node accesses readable.
        struct Node {
            data: QuadTreeData,
            parent_index: usize,
        }

        // Simulate a DFS using a stack.
        let mut stack = VecDeque::from([(0, 0)]);
        let mut nodes = Vec::new();

        while let Some((parent_index, depth)) = stack.pop_back() {
            let max = FromBytes::from_bytes(byte_stream).trace::<Self>()?;
            let min = FromBytes::from_bytes(byte_stream).trace::<Self>()?;
            let half_size = FromBytes::from_bytes(byte_stream).trace::<Self>()?;
            let center = FromBytes::from_bytes(byte_stream).trace::<Self>()?;

            let children = match depth < MAX_DEPTH {
                true => Vec::with_capacity(CHILD_COUNT),
                false => Vec::new(),
            };

            // Add next node lookups to the stack.
            if depth < MAX_DEPTH {
                let self_index = nodes.len();

                for _counter in 0..CHILD_COUNT {
                    stack.push_back((self_index, depth + 1));
                }
            }

            let data = QuadTreeData {
                max,
                min,
                half_size,
                center,
                children,
            };
            nodes.push(Node { data, parent_index });
        }

        // Go back to front and assign nodes to their parents. Node 0 is the root
        // node so we skip that.
        while nodes.len() > 1 {
            let node = nodes.pop().unwrap();
            nodes[node.parent_index].data.children.insert(0, node.data);
        }

        let root_node = nodes.pop().unwrap().data;
        Ok(root_node)
```

Notable differences:
- Went back to using a `Vec` for the nodes since `HashMap` is not a good data structure for this use case
- Use the same storage for the nodes and the parent indices since they have a 1 to 1 mapping (introduced a helper struct `Node` to help with keeping this structured)
- Using `insert(0)` instead of `push()` followed by `reverse`
- Let Rust infer the types (`stack`, `min`, `max`, `children`, etc.). Rust has very good type inference, let it do the work!

I have confirmed through testing that this current version creates the exact same quad tree as your code.

ragnarok_formats/src/map.rs
```diff
@@ -55,6 +72,79 @@ bitflags::bitflags! {
         const CLIFF = 0b00001000;
     }
 }
+// An internal dfs function needs generics, doesn't know how to put using dfs.
+// TODO: find a easier way to parse QuadTreeData
```
vE5li, line 76: These kind of comments are not really helpful, I would prefer if we just removed them.

ragnarok_formats/src/map.rs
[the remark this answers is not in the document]
vE5li: If there is more stuff to be done I would prefer you put a good `TODO` comment there.

E.g. `// TODO: Parse remaining fields`

Also the `// TODO: verify version` is now obsolete and can be removed

ragnarok_formats/src/map.rs
kokosha: I was thinking that there are no more simple solution to the problem using stack and would probably need to implement using recursive function to generate QuadTree. But this also means that the recursion function will need to derive all the function generics of the parent function and the code would be a lot more messier.

ragnarok_formats/src/map.rs
kokosha: Oh, okay, it is much more neat than I expected. With the modification that you made, the code becomes more neat.
I need to process if everything is alright, because the nodes vector, may be not using the correct index. Need to
confirm by hand with the program running. Later, I will return if the code is working and commit the code.

> Let Rust infer the types (`stack`, `min`, `max`, `children`, etc.). Rust has very good type inference, let it do the work!
I was trying to print the elements and was not detecting the type after `FromBytes::from_bytes(byte_stream).trace::<Self>()?;` doesn't know the reason, but I will remember to remove in the next commit.

> Using insert(0) instead of push() followed by reverse
Strategy 1. Using this method of insert to begin
When you insert to begin, you shift all elements from vector by 1 and place the element. 
For insertion, you use 4 operation. For the shift all elements from vector by, you would use (1+2+3) operation 6 operation. Resulting in 10 operation
Strategy 2. Using this method of insert to end and reverse, you only shift during reverse.
For insertion, you use 4 operation. And for reverse, you use more 4 operation. Resulting in 8 operations.

For code understanding, the strategy 1 will be easier to understand.

ragnarok_formats/src/map.rs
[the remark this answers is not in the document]
kokosha: Ok, I will make these changes.

ragnarok_formats/src/map.rs
vE5li: That's fine, but the comment doesn't really have anything to do with that

kokosha:
Implemented the correction, need to see if the comment of reverse is ok

PR #98, opened 2024-09-28
Implement the bindless pattern for the geometry.

hasenbanck:
To prepare the move of the graphics engine into it's own crate, I started to convert the geometry rendering to the bindless parttern. This greatly reduces the time to record the render pass data. We use the fact, that the lifetime of all geometry is bound to the lifetime of a map. We collect all textures and vertex data into one binding array / buffer, and can then draw all geometry of a map in one draw call. On the CPU side this reduced the time to render to around 1/4 and drawcall greatly on my system. The GPU render times only got slightly better (it seems for my 4080 the scene is "too simple" and we are bottlenecked by log occupancy).

I removed the caching for maps and models, since it interfered with the assembly process (I also think the caching for those two resources is not very usefull, since loading is very snappy already and caching would result in memory bloat).

I removed the picker geometry renderer, since it was unused anyhow. I added a tile renderer for the debugging of tiles, since I couldn't re-use the geometry renderer for it anymore.

I removed the debug marker for two function, so that the compiler can better optimize the tight loop for the writing of geometry instructions.


hasenbanck:
I will resolve the merge conflicts.

hasenbanck:
I rebased it and the branch is feature complete now. But I see some stalls in the pointer shadow renderer which I want to investigate in the comming week.

hasenbanck:
Just want to add that we see these spiked on both the main branch and this branch. So the review could start non the less.

hasenbanck:
Okay, fixed all problem:

- Bindless on DX12 works (workaround applies until WGPU fixed the bug)
- Point shadow renders correctly using bindless.

The performance is not as good as I expected and we seem to have a lot of spices. I think this is a result of the way we currently do multi-threading. A test running all render branches on the main thread resultet in better performance.

hasenbanck:
Rebased the branch again against main.

hasenbanck:
I'm currently reworking all renderer to use the bindless approach and seperate "what" to paint and "how" to paint. I will keep open this PR for reference, but won't rebase it. The next PR, which I hope to finish this weekend will replace this PR instead.

PR #102, opened 2024-09-29
Adding Resurrection to the Player

kokosha:
The resurrection of character caused some bugs with the client. 
Fixing when the mouse goes out of the range of tile render walk indicator.


korangar/src/loaders/action/mod.rs
```diff
@@ -49,6 +49,13 @@ impl AnimationState {
         self.factor = Some(movement_speed as f32 * 100.0 / 150.0);
     }
 
+    pub fn death(&mut self, client_tick: ClientTick) {
```
vE5li, line 52: Something like `dead` is more in like with the other functions on this struct

korangar/src/main.rs
```diff
@@ -743,7 +747,12 @@ fn main() {
                             entities.push(npc);
                         }
                         NetworkEvent::RemoveEntity(entity_id) => {
-                            entities.retain(|entity| entity.get_entity_id() != entity_id);
+                            if entities[0].get_entity_id() == entity_id {
+                                entities[0].set_death(client_tick);
```
vE5li, line 751: Same here, `set_dead`, instead of `set_death`

korangar/src/world/map/mod.rs
```diff
@@ -341,6 +341,10 @@ impl Map {
         const OFFSET: f32 = 1.0;
         const TILE_SIZE: f32 = 5.0;
 
+        if !(0 <= position.x && position.x < self.width && 0 <= position.y && position.y < self.height) {
```
vE5li, line 344: This can just be re-written as
```rust
if position.x >= self.width || position.y >= self.height {
```
since `position.x` and `position.y` are `usize`.

Looking at the code I am unsure why this check is necessary, since we only ever call `render_walk_indicator` if the case that the mouse is hovering a tile. What was your reasoning here?

korangar/src/main.rs
```diff
@@ -724,6 +724,10 @@ fn main() {
                         NetworkEvent::CharacterSlotSwitchFailed => {
                             interface.open_window(&application, &mut focus_state, &ErrorWindow::new("Failed to switch character slots".to_owned()));
                         },
+                        NetworkEvent::ResurrectPlayer(data) => {
+                            entities[0].set_idle(client_tick);
+                            interface.close_window_with_class(&mut focus_state, RespawnWindow::WINDOW_CLASS);
+                        }
```
vE5li, line 730: Seem like `data` here is actually the `player_id`.

Also, is this packet sent for other players too? If so, we should select the correct one from the list to resurrect and only close the window if the correct player (ourselves) was resurrected.

ragnarok_packets/src/lib.rs
```diff
@@ -1460,6 +1460,13 @@ pub struct MovingEntityAppearedPacket {
     #[length(24)]
     pub name: String,
 }
```
vE5li, line 1462: Nit: there should be an empty line here

ragnarok_packets/src/lib.rs
```diff
@@ -1460,6 +1460,13 @@ pub struct MovingEntityAppearedPacket {
...
+#[derive(Debug, Clone, Packet, ServerPacket, MapServer)]
+#[cfg_attr(feature = "interface", derive(korangar_interface::elements::PrototypeElement))]
+#[header(0x0148)]
+pub struct ResurrectionPacket {
+    pub entity_id: EntityId,
+    pub p_type: u16,
```
vE5li, line 1468: What is `p_type`?

ragnarok_packets/src/lib.rs
```diff
@@ -1793,6 +1800,21 @@ pub struct ObjectiveDetails1 {
...
+pub struct NotificationPacket {
+    pub entity_id: EntityId,
+    pub target_id: u32,
+    pub server_tick: u32,
+    pub src_speed: u32,
+    pub dmg_speed: u32,
+    pub damage: i16,
+    pub div: u16,
+    pub p_type: u8,
+    pub damage2: i16,
```
vE5li, line 1815: We try to avoid abbreviation in the code base, so `src_speed` should be `source_speed` etc.

Also wondering what `p_type` and `damage2` are in this context.

ragnarok_packets/src/lib.rs
```diff
@@ -1793,6 +1800,21 @@ pub struct ObjectiveDetails1 {
     pub mob_name: String,
 }
 
+#[derive(Debug, Clone, Packet, ServerPacket, MapServer)]
+#[cfg_attr(feature = "interface", derive(korangar_interface::elements::PrototypeElement))]
+#[header(0x008A)]
+pub struct NotificationPacket {
```
vE5li, line 1806: Where did you get the name `NotificationPacket`?

vE5li (commented):
I tested the changes and it works like a charm :+1:

I just have some code related questions.

korangar/src/world/map/mod.rs
kokosha: There is a bug that the game crash in get_tile(), when you enter Izlude airport and try to enter the flying ship, with cursor above the teleporting portal.

ragnarok_packets/src/lib.rs
kokosha: the packet type, i search rathena source code and the only possible value is 0.

korangar/src/main.rs
kokosha: i will change

korangar/src/loaders/action/mod.rs
kokosha: i will change

ragnarok_packets/src/lib.rs
kokosha: I don't know what this packet is about, is a placeholder name.

ragnarok_packets/src/lib.rs
kokosha: I will change to remove abbreviation

korangar/src/main.rs
kokosha: There are more stuff to handle... There is the need to handle for other player dead and revival also.

ragnarok_packets/src/lib.rs
vE5li: Then please rename the field to `packet_type` and set the `#[new_default]` attribute. A comment saying `/// Alywas 0 on rAthena` would also be appreciated

vE5li:
I had some fun improving the network packets, so I took the liberty of fixing up the rest of the PR as well.

PR #104, opened 2024-10-02
Experimental SDL2 port.

hasenbanck:
This is an experimental port to SDL2 that was created to test, if this fixed the latency issues some users were facing.


hasenbanck:
Switching to SDL2 didn't improve the percieved latency issues the users were facing. So there is no reason to switch from winit to SDL2.

PR #112, opened 2024-10-06
Picker values can fully store 32-bit values.

hasenbanck:
This removes the implicit reliance of an entity ID to not use the full 32-bit value range. Since we can't assume the actual range that is used inside the 32 bit of an EntityId, we need to properly encode the full 32 bit inside the picker buffer.

For this we use a Rg32Uint texture, to store the picker values. This doubles the picker texture size (4K texture is around 65 MiB). The runtime of the shader is still very low and insignificant compared to other render passes.

This also allows us to encode much more data into the picker value later, if we need to do so.

### DIFF
--- a/korangar/src/graphics/renderers/picker/geometry/geometry.wgsl
+++ b/korangar/src/graphics/renderers/picker/geometry/geometry.wgsl
@@ -35,12 +35,12 @@ fn vs_main(
 fn fs_main(
     @location(0) texture_coordinates: vec2<f32>,
     @location(1) texture_index: i32
-) -> @location(0) u32 {
+) -> @location(0) vec2<u32> {
     let diffuse_color = textureSample(textures[texture_index], texture_sampler, texture_coordinates);
 
     if (diffuse_color.a != 1.0) {
         discard;
     }
 
-    return 0u;
+    return vec2<u32>(0u);
 }
--- a/korangar/src/graphics/renderers/picker/selector/selector.wgsl
+++ b/korangar/src/graphics/renderers/picker/selector/selector.wgsl
@@ -9,6 +9,7 @@ var<push_constant> constants: Constants;
 @compute
 @workgroup_size(1)
 fn cs_main() {
-    let texel_value = textureLoad(texture, constants.pointer_position, 0);
-    buffer[0] = texel_value.x;
+    let texel_value = textureLoad(texture, constants.pointer_position, 0).rg;
+    buffer[0] = texel_value.r;
+    buffer[1] = texel_value.g;
 }
--- a/korangar/src/graphics/renderers/picker/target.rs
+++ b/korangar/src/graphics/renderers/picker/target.rs
@@ -5,33 +5,32 @@ use ragnarok_packets::EntityId;
 #[cfg(feature = "debug")]
 use crate::world::MarkerIdentifier;
 
-#[cfg(feature = "debug")]
-enum MarkerEncoding {
-    Object = 10,
-    LightSource = 11,
-    SoundSource = 12,
-    EffectSource = 13,
-    Entity = 14,
-    Shadow = 15,
+#[repr(u32)]
+pub(super) enum PickerValueType {
+    Nothing,
+    Tile,
+    Entity,
+    #[cfg(feature = "debug")]
+    ObjectMarker,
+    #[cfg(feature = "debug")]
+    LightSourceMarker,
+    #[cfg(feature = "debug")]
+    SoundSourceMarker,
+    #[cfg(feature = "debug")]
+    EffectSourceMarker,
+    #[cfg(feature = "debug")]
+    EntityMarker,
+    #[cfg(feature = "debug")]
+    ShadowMarker,
 }
 
-/// Encoding a `PickerTarget` as `u32` has the following format:
+/// Encoding of a `PickerTarget` as `u64` has the following format:
 ///
-/// If the most significant bit is set (`(1 << 31)`), it encodes a position. In
-/// that case, the next 15 bits are used to store the X position, and the
-/// remaining 16 bits contain the Y position. That gives us an effective maximum
-/// value of 16384 for the X channel, which for Ragnarok Online is plenty.
-///
-/// If the `debug` feature is enabled, the most significant byte can contain a
-/// value matching that of a `MarkerEncoding` (private enum). If that is the
-/// case, the remaining 3 bytes are to be interpreted as the value of the
-/// marker.
-///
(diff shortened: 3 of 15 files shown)


korangar/src/graphics/renderers/picker/geometry/geometry.wgsl
vE5li, line 1: I thought we removed the picker geometry renderer. Is therer a reason we keep this?

korangar/src/graphics/renderers/picker/selector/selector.wgsl
```diff
@@ -9,6 +9,7 @@ var<push_constant> constants: Constants;
 @compute
 @workgroup_size(1)
 fn cs_main() {
-    let texel_value = textureLoad(texture, constants.pointer_position, 0);
-    buffer[0] = texel_value.x;
+    let texel_value = textureLoad(texture, constants.pointer_position, 0).rg;
+    buffer[0] = texel_value.r;
+    buffer[1] = texel_value.g;
```
vE5li, line 14: Can we make the `buffer` an `array<u64>`?

It feels a bit wrong that the buffer is of type `u64` but the shader treats it as `u32`

korangar/src/graphics/renderers/picker/target.rs
```diff
@@ -5,32 +5,29 @@ use ragnarok_packets::EntityId;
...
+    #[cfg(feature = "debug")]
+    LightSourceMarker = !0 - 1,
+    #[cfg(feature = "debug")]
+    SoundSourceMarker = !0 - 2,
+    #[cfg(feature = "debug")]
+    EffectSourceMarker = !0 - 3,
+    #[cfg(feature = "debug")]
+    EntityMarker = !0 - 4,
+    #[cfg(feature = "debug")]
+    ShadowMarker = !0 - 5,
```
vE5li, line 23: Just out of curiosity, why did you decide to fill the `u32` with `1`s for the markers?

I think at this point I would set `Tile = 1` and let the other enum values be inferred.

korangar/src/graphics/renderers/picker/tile/title.wgsl
```diff
@@ -21,6 +21,6 @@ fn vs_main(
 }
 
 @fragment
-fn fs_main(@location(0) @interpolate(flat) identifier: u32) -> @location(0) u32 {
-    return identifier;
+fn fs_main(@location(0) @interpolate(flat) identifier: u32) -> @location(0) vec2<u32> {
+    return vec2<u32>(identifier, 1);
```
vE5li, line 25: I am sort of only half happy with this magic number here. Maybe we should add a comment?

korangar/src/graphics/renderers/picker/geometry/geometry.wgsl
hasenbanck: We haven't removed it yet. #98 is the PR that removes it.

korangar/src/graphics/renderers/picker/selector/selector.wgsl
hasenbanck: WGSL has no u64.

https://www.w3.org/TR/WGSL/#plain-types-section

korangar/src/graphics/renderers/picker/tile/title.wgsl
hasenbanck: We could maybe remove this magic value by using a pipeline specialization constant? Should I try this approach? Then we could use the enum value directly with no magic constants.

korangar/src/graphics/renderers/picker/geometry/geometry.wgsl
vE5li: Sorry, my bad!

korangar/src/graphics/renderers/picker/tile/title.wgsl
vE5li: That would be super nice actaully

korangar/src/graphics/renderers/picker/tile/title.wgsl
vE5li, line 1: I just realized this file has a typo in it. Can you fix it?

korangar/src/graphics/renderers/picker/target.rs
vE5li: Are there other places than the tests (and the tile shader that we want to use specialization constants for now)?

If it's only the tests I would prefer having to fix them over giving the variants arbitrary values. On that note, we can also start at `0` now right? So we don't even need `Tile = 1`

korangar/src/graphics/renderers/picker/tile/title.wgsl
hasenbanck: I implement it later this evening.

korangar/src/graphics/renderers/picker/tile/title.wgsl
hasenbanck: Sure will do.

korangar/src/graphics/renderers/picker/target.rs
hasenbanck: I think the tests are the last place. If we make Tile = 0, what would the clear value? I mean, we need something that means "nothing selected".

(High 0, Low 0) would then mean "Tile at X=0, Y=0", right?

hasenbanck:
Add changes based on your feedback:

1. enum PickerValueType's value is not auto assigned. But I added a "Nothing" type to signal, that nothing was selected with the mouse. It's assigned on fall through, so everything that isn't the other enum values, is automatically "Nothing".
2. Removed magic constant by setting it with the help of a pipeline constant.
3. Added funky linter warning by replacing a cast with a concrete conversion to u64.
4. Updated tests.

Feel free to change stuff you don't like.

hasenbanck:
Still thinkint about the clear value. We could do the following when creatiing a new render pass for the PickerRenderTarget:

```rust
        let (clear_high, clear_low) = <(u32, u32)>::from(PickerTarget::Nothing);
        let clear_color = wgpu::Color {
            r: f64::from(clear_high),
            g: f64::from(clear_low),
            b: 0.0,
            a: 0.0,
        };
````

korangar/src/graphics/renderers/picker/target.rs
hasenbanck: We check those implicitly by using it as a fall through case. But in the old branch, we didn't even gave the clear value a linked enum value. The current branch does this now. So if a value inside the picker texture has the clear value or any other unknown type value, it will be interpreted as "Nothing".

vE5li:
Very nice, I really like this :+1:

PR #116, opened 2024-10-13
Fix RSM

kokosha:
Fix some calculation in RSM rendering


vE5li:
Very nice! The game looks much better already :)

I made some small adjustments to your PR, notably:
- Remove `temp_matrix` and use `transform_matrix` instead, since it was always an identity matrix in the beginning
- Moved the version check to the `load` method so that we can load a fallback model in case of an unsupported version
- (_Minor_) Use `for_each` for `calculate_transformation_matrix` on the child nodes

PR #119, opened 2024-10-22
Implement and use a texture atlas

hasenbanck:
I initially created this for Metal and limited backends, but it seems a good optimization non the less. So it's right now used for all code paths, which should keep the codebase simpler and easier to maintain.

I initially used dynamic uniform buffers in for the model drawer, but in the end didn't needed to do that. I still include the "DynamicUniformBuffer" struct, since it cleans up some other code and might be helpful for future features.

I sepperated "AttachmentTexture" and "Textures", to make clear that they are different things. Textures now have their own bind group, which should help later with Metal and limited backends.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1471,6 +1471,7 @@ version = "0.1.0"
 dependencies = [
  "cgmath",
  "derive-new",
+ "image",
  "korangar_interface",
  "rand",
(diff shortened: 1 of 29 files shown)


vE5li:
Very nice improvement :+1:

PR #128, opened 2024-10-28
Even faster way to do cone/sphere intersections

hasenbanck:
This is now optimized to the extreme. We still have one 3x3 matrix operation in the hot loop, but modern GPU have fast paths for those. The intersection test itself is now very, very efficient. I doubt there is something more efficient out there. We take full
advantage of the fact, that our cone is axis aligned and also the origin of the space.


korangar/src/graphics/passes/light_culling/shader/light_culling.wgsl
```diff
@@ -152,6 +165,20 @@ fn calculate_tile_vectors(tile_x: u32, tile_y: u32) -> mat4x3<f32> {
...
+fn create_cone_rotation_matrix(cone_axis: vec3<f32>) -> mat3x3<f32> {
+    let z = normalize(cone_axis);
+    let temp = select(vec3<f32>(1.0, 0.0, 0.0), vec3<f32>(0.0, 1.0, 0.0), abs(z.x) > 0.9);
+    let x = normalize(cross(temp, z));
+    let y = cross(z, x);
+    return mat3x3<f32>(
+        x.x, y.x, z.x,
+        x.y, y.y, z.y,
+        x.z, y.z, z.z
+    );
```
vE5li, line 179: I am not super happy with the naming here. Isn't what you called `x`, `y` and `z` here usually `row0`, `row1` and `row2` (or column, I don't matrix very often)?

Also rename `temp` please, at the very least it shoudn't be abbreviated.

hasenbanck: Added more context to make clear, what the vectors are meaning in this context (they are essentially the new x,y,z axis!).

PR #130, opened 2024-10-29
Render all parts of entities

kokosha:


korangar/src/loaders/animation/mod.rs
```diff
@@ -0,0 +1,474 @@
...
+            let position;
+            let animation_index = framepart.animation_index as usize;
+            let sprite_number = framepart.sprite_number as usize;
+            texture = &self.animation_pair[animation_index].sprites.textures[sprite_number];
+            position = Vector2::new(
+                animation.frames[0].offset.x as f32,
+                animation.frames[0].offset.y as f32 + ((animation.frames[time].size.y - 1) / 2) as f32,
+            ) / 10.0;
+
+            // Generate vertex
```
hasenbanck, line 442: You are creating the vertex position of the sprite here. Couldn't we instead create a transformation matrix instead, which we then apply on a unit square inside the shader?

korangar/src/loaders/animation/mod.rs
```diff
@@ -0,0 +1,474 @@
...
+        }
+
+        let mut datas = Vec::new();
+        for framepart in frame.frameparts.iter() {
+            let texture;
+            let position;
+            let animation_index = framepart.animation_index as usize;
+            let sprite_number = framepart.sprite_number as usize;
+            texture = &self.animation_pair[animation_index].sprites.textures[sprite_number];
+            position = Vector2::new(
```
hasenbanck, line 437: It would be great if you could explain the operation here. Especially the use "/2" and "/10" parts.

korangar/src/loaders/animation/mod.rs
```diff
@@ -0,0 +1,474 @@
...
+#[derive(Clone, PrototypeElement)]
+pub struct AnimationData {
+    pub animation_pair: Vec<AnimationPair>,
+    pub animations: Vec<Animation>,
+    pub delays: Vec<f32>,
+    #[hidden_element]
+    pub entity_type: EntityType,
+}
+
+// This function generates the convertion to the big rectangle
```
hasenbanck, line 388: This function looks like a screen space to NDC conversion! What is "big rectangle" in this context? Can you please explain this more?

korangar/src/loaders/animation/mod.rs
```diff
@@ -0,0 +1,474 @@
...
+        let mut new_height = frame_y.upleft.y + frame_y.size.y as i32;
+        new_width -= upleft_x;
+        new_height -= upleft_y;
+
+        let mut new_frameparts = Vec::<FramePart>::new();
+        for index in 0..frames.len() {
+            new_frameparts.append(&mut frames[index].frameparts);
+        }
+
+        /* As the origin is (0,0), you get the upleft point of the retangle and
```
hasenbanck, line 354: We normally use line comments for multi line comments too. Here are also some typos.

korangar/src/loaders/animation/mod.rs
```diff
@@ -0,0 +1,474 @@
...
+        let frame_y = frames.iter().max_by_key(|frame| frame.upleft.y + frame.size.y as i32).unwrap();
+
+        // Calculate the new rectangle that is formed
+        let mut new_width = frame_x.upleft.x + frame_x.size.x as i32;
+        let mut new_height = frame_y.upleft.y + frame_y.size.y as i32;
+        new_width -= upleft_x;
+        new_height -= upleft_y;
+
+        let mut new_frameparts = Vec::<FramePart>::new();
+        for index in 0..frames.len() {
```
hasenbanck, line 350:
```
        for frame in frames.iter_mut() {
            new_frameparts.append(&mut frame.frameparts);
        }
```

korangar/src/loaders/animation/mod.rs
```diff
@@ -0,0 +1,474 @@
...
+        // Find the upmost and leftmost coordinates
+        let upleft_x = frames.iter().min_by_key(|frame| frame.upleft.x).unwrap().upleft.x;
+        let upleft_y = frames.iter().min_by_key(|frame| frame.upleft.y).unwrap().upleft.y;
+
+        // Find the downmost and rightmost coordinates
+        let frame_x = frames.iter().max_by_key(|frame| frame.upleft.x + frame.size.x as i32).unwrap();
+        let frame_y = frames.iter().max_by_key(|frame| frame.upleft.y + frame.size.y as i32).unwrap();
+
+        // Calculate the new rectangle that is formed
+        let mut new_width = frame_x.upleft.x + frame_x.size.x as i32;
```
hasenbanck, line 344: Unneeded conversion.

korangar/src/loaders/animation/mod.rs
```diff
@@ -0,0 +1,474 @@
...
+                frameparts: vec![frame_part],
+            };
+            return frame;
+        }
+        // Find the upmost and leftmost coordinates
+        let upleft_x = frames.iter().min_by_key(|frame| frame.upleft.x).unwrap().upleft.x;
+        let upleft_y = frames.iter().min_by_key(|frame| frame.upleft.y).unwrap().upleft.y;
+
+        // Find the downmost and rightmost coordinates
+        let frame_x = frames.iter().max_by_key(|frame| frame.upleft.x + frame.size.x as i32).unwrap();
```
hasenbanck, line 340: Unneeded conversion.

korangar/src/loaders/animation/mod.rs
```diff
@@ -0,0 +1,474 @@
...
+pub struct Frame {
+    pub offset: Vector2<i32>,
+    pub upleft: Vector2<i32>,
+    pub size: Vector2<i32>,
+    pub remove_offset: Vector2<i32>, // Is only used for final shift
+    pub frameparts: Vec<FramePart>,
+}
+
+impl Frame {
+    // This function generate new rame
```
hasenbanck, line 302: Typo. You can also use "///" to make a function documentation instead.

korangar/src/loaders/animation/mod.rs
```diff
@@ -0,0 +1,474 @@
...
+    pub fn merge_frame_part(frames: &mut Vec<Frame>) -> Frame {
```
hasenbanck, line 303: You don't seem to change the Vec length here. So you can use `&mut[]` instead.

korangar/src/loaders/animation/mod.rs
```diff
@@ -0,0 +1,474 @@
...
+        }
+    }
+}
+
+#[derive(Clone)]
+pub struct FramePart {
+    pub animation_index: i32,
+    pub sprite_number: i32,
+    pub offset: Vector2<i32>,
+    pub upleft: Vector2<i32>, // Is only used for internal calculation
```
hasenbanck, line 285: Never used.

korangar/src/loaders/animation/mod.rs
```diff
@@ -0,0 +1,474 @@
...
+    }
+
+    pub fn get(
+        &mut self,
+        sprite_loader: &mut SpriteLoader,
+        action_loader: &mut ActionLoader,
+        entity_filename: Vec<String>,
+        entity_type: EntityType,
+    ) -> Result<AnimationData, LoadError> {
+        let hash = match entity_type {
```
hasenbanck, line 266: I think it's better to create a enum here, that we then can use as a hash instead (an enum that has fields for each type). Encoded strings are very brittle.

korangar/src/loaders/animation/mod.rs
```diff
@@ -0,0 +1,474 @@
...
+                frame.offset.x += max_offset_x - min_offset_x + 2;
+                frame.offset.y += max_offset_y - min_offset_y + 2;
+                frame.size = Vector2::new(new_width, new_height);
+                frame.remove_offset.x = max_offset_x - min_offset_x + 2;
+                frame.remove_offset.y = max_offset_y - min_offset_y + 2;
+            });
+            animations.push(Animation { frames });
+        }
+        let animation_data = AnimationData {
+            delays: (&vec[0]).actions.delays.clone(),
```
hasenbanck, line 244: "vec" is already borrowed.

korangar/src/loaders/animation/mod.rs
```diff
@@ -0,0 +1,474 @@
...
+                        continue;
+                    }
+                    generate.push(animations_list[animation_pair_index][action_index][motion_index].clone());
+                }
+                let frame = Frame::merge_frame_part(&mut generate);
+
+                frames.push(frame);
+            }
+
+            // The problem is that the offset is not in the correct proportion and
```
hasenbanck, line 201: Typo. Please also use proper punctuation here.

korangar/src/loaders/animation/mod.rs
```diff
@@ -0,0 +1,474 @@
...
+
+        // Generate for each action, the several motions that it have and create a
+        // offset of retangles.
+        for action_index in 0..action_size {
+            let motion_size = vec[0].actions.actions[action_index].motions.len();
+            let mut frames: Vec<Frame> = Vec::new();
+            for motion_index in 0..motion_size {
+                let mut generate: Vec<Frame> = Vec::new();
+
+                for animation_pair_index in 0..animation_pair_size {
```
hasenbanck, line 187: No need to iter using the index, you can iter over animations_list directly.

korangar/src/loaders/animation/mod.rs
```diff
@@ -0,0 +1,474 @@
...
+            }
+            animations_list.push(animation_frames);
+            animation_index += 1;
+        }
+        let action_size = vec[0].actions.actions.len();
+        let animation_pair_size = vec.len();
+
+        let mut animations: Vec<Animation> = Vec::new();
+
+        // Generate for each action, the several motions that it have and create a
```
hasenbanck, line 179: typos

korangar/src/loaders/animation/mod.rs
```diff
@@ -0,0 +1,474 @@
...
+                                blue: 0.0,
+                                alpha: 0.0,
+                            },
+                        };
+
+                        // Scale the image
+                        // Try to match the first type of zoom, if doesn't match find the second method
+                        let zoom = match motion.sprite_clips[pos].zoom {
+                            Some(value) => (value, value).into(),
+                            None => match motion.sprite_clips[pos].zoom2 {
```
hasenbanck, line 105:
```
        None => motion.sprite_clips[pos].zoom2.unwrap_or_else(|| (1.0, 1.0).into()),
```

korangar/src/loaders/animation/mod.rs
```diff
@@ -0,0 +1,474 @@
...
+                        if sprite_number == -1 {
+                            continue;
+                        }
+                        let texture_size = animation_pair.sprites.textures[sprite_number as usize].get_size();
+                        let mut height = texture_size.height;
+                        let mut width = texture_size.width;
+                        // Apply color filter in the image
+                        let color = match motion.sprite_clips[pos].color {
+                            Some(color) => {
+                                let alpha = ((((color >> 24) & 0xFF) as u8) as f32 / 255.0) as f32;
```
hasenbanck, line 86: Unnecessary cast here.

korangar/src/loaders/animation/mod.rs
```diff
@@ -0,0 +1,474 @@
...
+            for action in animation_pair.actions.actions.iter() {
+                motion_index = 0;
+                let mut action_frames: Vec<Frame> = Vec::new();
+                for motion in action.motions.iter() {
+                    let mut motion_frames: Vec<Frame> = Vec::new();
+                    if motion.sprite_clip_count == 0 {
+                        continue;
+                    }
+                    for position in 0..motion.sprite_clip_count {
+                        let frame_part: FramePart;
```
hasenbanck, line 74: Move the declaration and initialization together.

korangar/src/loaders/animation/mod.rs
```diff
@@ -0,0 +1,474 @@
...
+            .collect();
+
+        // The sprite is stored as rgba in animation_pair.sprites.rgba_images or
+        // animation_pair.sprites.palette_images
+        // The sprite is stored as texture in animation_pair.sprites.textures
+
+        // For each animation, we collect all the framepart need to generate the frame
+        let mut animations_list: Vec<Vec<Vec<Frame>>> = Vec::new();
+        let mut animation_index: usize = 0;
+        let mut action_index: usize = 0;
```
hasenbanck, line 54: I have a feeling that the following two index are in reality an enumeration values that you could get my using "enumerate()" on an iter?

korangar/src/loaders/animation/mod.rs
```diff
@@ -0,0 +1,474 @@
...
+impl AnimationLoader {
+    pub fn load(
+        &mut self,
+        sprite_loader: &mut SpriteLoader,
+        action_loader: &mut ActionLoader,
+        entity_filename: Vec<String>,
+        entity_type: EntityType,
+    ) -> Result<AnimationData, LoadError> {
+        // Create animation pair with sprite and action
+        let vec: Vec<AnimationPair> = entity_filename
```
hasenbanck, line 39: "vec" naming here is very confusing. Better find a more descriptive name.

korangar/src/loaders/animation/mod.rs
```diff
@@ -0,0 +1,474 @@
...
+#[derive(new)]
+pub struct AnimationLoader {
+    device: Arc<Device>,
+    queue: Arc<Queue>,
+    #[new(default)]
+    // The string will be type of entity
+    // 0_{body_id}_{head_id}
+    // 1_{monster_id}
+    // 2_{npc_id}
+    cache: HashMap<String, AnimationData>,
```
hasenbanck, line 27: As I wrote below: Better find use an enum here as the key. You can also use hashbrown's hashmap here instead of the std once, since it's much faster.

korangar/src/loaders/animation/mod.rs
```diff
@@ -0,0 +1,474 @@
...
+use korangar_interface::elements::PrototypeElement;
+use num::Zero;
+use wgpu::{Device, Extent3d, Queue, TextureDescriptor, TextureDimension, TextureFormat, TextureUsages};
+
+use super::error::LoadError;
+use crate::graphics::Texture;
+use crate::loaders::{ActionLoader, Actions, AnimationState, Sprite, SpriteLoader};
+use crate::{Color, EntityType};
+
+// TODO: use cache later, the memory will be increase with this hashmap, until
```
hasenbanck, line 16: Yeah, you could use the ResourceCache here with a sensible limit instead.

korangar/src/graphics/instruction.rs
```diff
@@ -177,8 +177,15 @@ pub struct EntityInstruction {
     pub world: Matrix4<f32>,
     pub texture_position: Vector2<f32>,
     pub texture_size: Vector2<f32>,
+    pub texture_top_left: Vector2<f32>,
```
hasenbanck, line 180: Why do we still need texture_position and texture_size? Can't we just create the vertex here? (Or even better: Just a simple 4x4 transformation matrix?)

korangar/src/graphics/instruction.rs
```diff
@@ -177,8 +177,15 @@ pub struct EntityInstruction {
...
+    pub texture_bottom_left: Vector2<f32>,
+    pub texture_top_right: Vector2<f32>,
+    pub texture_bottom_right: Vector2<f32>,
+    pub depth_extra: f32,
```
hasenbanck, line 184: What is depth_extra? Can't merge in into depth_offset instead? (in linear space of course)

korangar/src/graphics/passes/point_shadow/entity.rs
```diff
@@ -23,9 +23,17 @@ const INITIAL_INSTRUCTION_SIZE: usize = 256;
...
+    texture_top_left: [f32; 2],
+    texture_bottom_left: [f32; 2],
+    texture_top_right: [f32; 2],
+    texture_bottom_right: [f32; 2],
     texture_position: [f32; 2],
     texture_size: [f32; 2],
     depth_offset: f32,
+    depth_extra: f32,
+    angle: f32,
+    foo: f32,
```
hasenbanck, line 35: If you need to use padding, then just use a padding as the last field.

```
padding: [u32;2]
```

You can then initialize this with Default::default()!

korangar/src/graphics/passes/point_shadow/shader/entity_bindless.wgsl
```diff
@@ -121,6 +138,50 @@ fn vertex_data(vertex_index: u32) -> Vertex {
     return Vertex(vec3<f32>(x, y, z), vec2<f32>(u, v), depth, curve);
 }
 
+fn vertex_data_new(vertex_index: u32, instance_index: u32) -> Vertex {
```
hasenbanck, line 141: Uff, non-uniform branching is not good. Can you try to make this branchless? (By using a truth table and maybe by using the "select" function?) You can also remove the old vertex data function then.

korangar/src/graphics/passes/point_shadow/shader/entity_bindless.wgsl
```diff
@@ -64,12 +73,20 @@ fn vs_main(
...
     output.original_curvature = instance.curvature;
     output.texture_index = instance.texture_index;
+    output.angle = instance.angle;
     return output;
 }
 
 @fragment
 fn fs_main(input: VertexOutput) -> @builtin(frag_depth) f32 {
-    let diffuse_color = textureSample(textures[input.texture_index], nearest_sampler, input.texture_coordinates);
+    var new_input = input.texture_coordinates;
```
hasenbanck, line 82: You need to document here, what and why you are doing here.

korangar/src/graphics/passes/point_shadow/shader/entity_bindless.wgsl
```diff
@@ -6,9 +6,17 @@ struct PassUniforms {
...
+    texture_top_left: vec2<f32>,
+    texture_bottom_left: vec2<f32>,
+    texture_top_right: vec2<f32>,
+    texture_bottom_right: vec2<f32>,
     texture_position: vec2<f32>,
     texture_size: vec2<f32>,
     depth_offset: f32,
+    depth_extra: f32,
+    angle: f32,
+    foo: f32,
```
hasenbanck, line 18: If the padding is the last field, you don't need to define it on the shader side.

korangar/src/graphics/passes/point_shadow/shader/entity.wgsl
```diff
@@ -119,6 +136,50 @@ fn vertex_data(vertex_index: u32) -> Vertex {
     return Vertex(vec3<f32>(x, y, z), vec2<f32>(u, v), depth, curve);
 }
 
+fn vertex_data_new(vertex_index: u32, instance_index: u32) -> Vertex {
```
hasenbanck, line 139: Dito as in the bindless point shader file. I won't repeat this for the other shaders.

korangar/src/graphics/passes/point_shadow/shader/entity.wgsl
```diff
@@ -62,12 +71,20 @@ fn vs_main(
...
     output.original_depth_offset = instance.depth_offset;
     output.original_curvature = instance.curvature;
+    output.angle = instance.angle;
     return output;
 }
 
 @fragment
 fn fs_main(input: VertexOutput) -> @builtin(frag_depth) f32 {
-    let diffuse_color = textureSample(texture, nearest_sampler, input.texture_coordinates);
+    var new_input = input.texture_coordinates;
```
hasenbanck, line 80: Dito as in the bindless point shader file. I won't repeat this for the other shaders.

korangar/src/graphics/passes/point_shadow/shader/entity.wgsl
```diff
@@ -6,9 +6,17 @@ struct PassUniforms {
...
+    texture_top_left: vec2<f32>,
+    texture_bottom_left: vec2<f32>,
+    texture_top_right: vec2<f32>,
+    texture_bottom_right: vec2<f32>,
     texture_position: vec2<f32>,
     texture_size: vec2<f32>,
     depth_offset: f32,
+    depth_extra: f32,
+    angle: f32,
+    foo: f32,
```
hasenbanck, line 18: Dito as in the bindless point shader file. I won't repeat this for the other shaders.

korangar/src/graphics/passes/forward/entity.rs
```diff
@@ -4,6 +4,7 @@ use std::sync::Arc;
 use bumpalo::Bump;
 use bytemuck::{Pod, Zeroable};
 use hashbrown::HashMap;
+use lunify::InstructionLayout;
```
hasenbanck, line 7: This seems to be an unused and wrong import.

korangar/src/graphics/passes/forward/shader/entity_bindless.wgsl
```diff
@@ -180,7 +203,7 @@ fn fs_main(input: VertexOutput) -> FragmentOutput {
     }
 
     var output: FragmentOutput;
-    output.fragment_color = vec4<f32>(final_color, 1.0);
+    output.fragment_color = vec4<f32>(final_color, final_alpha);
```
hasenbanck, line 206: You set the real alpha here. Should we also only disgard the fragment on depth == 0.0 ?

hasenbanck (changes requested):
I finished the fruist pass review. Are you using clippy? It showed me many lints. Some easy quick fixes!

korangar/src/loaders/sprite/mod.rs
```diff
@@ -20,6 +21,10 @@ use crate::loaders::GameFileLoader;
 pub struct Sprite {
     #[hidden_element]
     pub textures: Vec<Arc<Texture>>,
+    pub rgba_count: u16,
```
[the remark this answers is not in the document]
kokosha: It may be used later, but as we are not using I will remove all unused elements for now.

korangar/src/loaders/sprite/mod.rs
```diff
@@ -119,12 +133,50 @@ impl SpriteLoader {
             })
             .collect();
 
+        let rgba_images: Vec<_> = sprite_data
+            .rgba_image_data
+            .clone()
+            .into_iter()
+            .map(|image_data| {
+                let data: Vec<_> = image_data.data.chunks_exact(4).flat_map(|w| [w[3], w[2], w[1], w[0]]).collect();
```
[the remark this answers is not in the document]
kokosha: I will change w to pixel and set the Vec to full type, where it can't be safely remove

korangar/src/loaders/sprite/mod.rs
```diff
@@ -119,12 +133,50 @@ impl SpriteLoader {
...
+                RgbaImageData {
+                    width: image_data.width,
+                    height: image_data.height,
+                    data,
+                }
+            })
+            .collect();
+
+        let palette_images: Vec<_> = sprite_data
```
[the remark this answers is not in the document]
kokosha: Dito. Set the Vec to full type where it can't be safely remove, or remove

korangar/src/loaders/sprite/mod.rs
```diff
@@ -119,12 +133,50 @@ impl SpriteLoader {
...
+            .palette_image_data
+            .clone()
+            .into_iter()
+            .map(|image_data| {
+                // decode palette image data if necessary
+                let data: Vec<u8> = image_data
```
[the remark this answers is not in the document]
kokosha: Dito

korangar/src/loaders/sprite/mod.rs
```diff
@@ -77,7 +92,7 @@ impl SpriteLoader {
             [palette.red, palette.green, palette.blue, alpha]
         }
 
-        let palette_images = sprite_data.palette_image_data.into_iter().map(|image_data| {
+        let palette_images = sprite_data.palette_image_data.clone().into_iter().map(|image_data| {
```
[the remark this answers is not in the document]
kokosha: Changing this and the places after.

korangar/src/loaders/sprite/mod.rs
```diff
@@ -63,9 +68,19 @@ impl SpriteLoader {
         let palette = sprite_data.palette.unwrap(); // unwrap_or_default() as soon as i know what
         // the default palette is
 
-        let rgba_images/*: Vec<Arc<ImmutableImage>>*/ = sprite_data
+        let rgba_images: Vec<_> = sprite_data
```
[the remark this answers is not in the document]
kokosha: Dito.

korangar/src/loaders/sprite/mod.rs
```diff
@@ -63,9 +68,19 @@ impl SpriteLoader {
...
             .rgba_image_data
-            .into_iter();
+            .clone()
```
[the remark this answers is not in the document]
kokosha: Dito.

korangar/src/loaders/animation/mod.rs
```diff
@@ -0,0 +1,474 @@
...
+    return Vector2::<f32>::new(x, y);
+}
+
+impl AnimationData {
+    pub fn render(
+        &self,
+        animation_state: &AnimationState,
+        camera_direction: usize,
+        head_direction: usize,
+    ) -> Vec<(Arc<Texture>, Vec<Vector2<f32>>, Vector2<f32>, Vector2<i32>, f32, Color, bool)> {
```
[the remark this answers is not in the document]
kokosha: This is possible, I will change the code to make this render instruction precomputed.

korangar/src/loaders/animation/mod.rs
kokosha: I am going to look into it, to check if it is possible

korangar/src/loaders/animation/mod.rs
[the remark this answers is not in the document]
hasenbanck: Don't pre-compute the instructions! Only pre-compute stuff that is agnostic to the position etc. of the animation.

korangar/src/loaders/animation/mod.rs
```diff
@@ -0,0 +1,474 @@
...
+            let bottom_left = top_left + new_vector.y * Vector2::<i32>::unit_y();
+            let top_right = top_left + new_vector.x * Vector2::<i32>::unit_x();
+            let bottom_right = top_left + new_vector;
+
+            let texture_top_left = convert_coordinate(top_left, frame.size);
+            let texture_bottom_left = convert_coordinate(bottom_left, frame.size);
+            let texture_top_right = convert_coordinate(top_right, frame.size);
+            let texture_bottom_right = convert_coordinate(bottom_right, frame.size);
+
+            let mut texture_coordinates = Vec::new();
```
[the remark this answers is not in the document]
kokosha: Changed to the simple form

korangar/src/loaders/animation/mod.rs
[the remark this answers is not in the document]
kokosha: Precomputed outside the render part.

korangar/src/loaders/animation/mod.rs
```diff
@@ -0,0 +1,474 @@
...
+        // Remove Doridori animation from Player
+        if self.entity_type == EntityType::Player && animation_state.action == 0 {
+            frame = &animation.frames[0];
+        } else {
+            frame = &animation.frames[time];
+        }
+
+        let mut datas = Vec::new();
+        for framepart in frame.frameparts.iter() {
+            let texture;
```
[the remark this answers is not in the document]
kokosha: Fixed.

korangar/src/loaders/animation/mod.rs
kokosha: Explained in a better way

korangar/src/loaders/animation/mod.rs
kokosha: This is the conversion of "normalized" coordinates of a frame part inside the frame bounding box rectangle with vertex [-1, 0], [-1, 2], [1, 0], [1, 2]

korangar/src/loaders/animation/mod.rs
kokosha: Resolved

korangar/src/loaders/animation/mod.rs
kokosha: Fixed the conversion.

korangar/src/loaders/animation/mod.rs
kokosha: Dito.

korangar/src/loaders/animation/mod.rs
kokosha: Fixed.

korangar/src/loaders/animation/mod.rs
kokosha: Changed to &mut[Frame]

korangar/src/loaders/animation/mod.rs
kokosha: Resolved

korangar/src/loaders/animation/mod.rs
kokosha: I am trying to find ways to directly iterate, hasn't found any ways yet.

korangar/src/loaders/animation/mod.rs
kokosha: Fixed typos

korangar/src/loaders/animation/mod.rs
hasenbanck: Example:

```
                for pair in &animations_list[0..animation_pair_size] {
                    if pair.len() <= action_index {
                        continue;
                    }
                    if pair[action_index].len() <= motion_index {
                        continue;
                    }
                    generate.push(pair[action_index][motion_index].clone());
                }
```

But it seems you can iterate over the full animations_list too.

korangar/src/loaders/animation/mod.rs
kokosha: Fixed

korangar/src/loaders/animation/mod.rs
kokosha: Fixed

korangar/src/loaders/animation/mod.rs
kokosha: Fixed

korangar/src/loaders/animation/mod.rs
kokosha: Yeah, this enumerate is a great way to deal with indexes

kokosha: Fixed

korangar/src/loaders/animation/mod.rs
kokosha: Renamed it to animation_pairs

korangar/src/loaders/animation/mod.rs
kokosha: Changed

korangar/src/loaders/animation/mod.rs
kokosha: I am going to change the hash function later, for now we use the inefficient string, as the commit is to big. The npc and monster is have id with number 30k or less, for player we can offset by 30k and create a number for each player.

korangar/src/loaders/animation/mod.rs
kokosha: Nice catch

korangar/src/loaders/animation/mod.rs
hasenbanck: The change could be as simple as:

```
#[derive(Clone, Debug, Hash)]
enum AnimationKey {
    Player { body: String, head: String },
    Monster(String),
    Npc(String),
}
```

kokosha: I changed to use number instead, see if it is okay.

hasenbanck: Okay, I think I know what you aimed for here. But since you are using the EntityId now, you can't really cache the values anymore, since EntityID's are not guranteed to be unique over the lifetime of the game session!

That would mean, as soon as an entity despawns, you have to remove it's animation from the cache, or else we could have  a wrong cache hit.

If we want to cache the animations,  then we need something that makes them identifiable for that they are. For example: A sprite is identifiable using it's path inside the game files. I think your approach using the file names was already good, but only needed a little refinement. What would you need to re-load the same animation again? I thought the filenames was already a good fit.

korangar/src/loaders/animation/mod.rs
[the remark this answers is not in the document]
hasenbanck: We are still returning a Vec<T> here! We then takr that list and later create the instructions using it.:

```
let datas = self
            .animation_data
            .render(&self.animation_state, camera_direction, self.head_direction);
```

You  can skip that and create the instructions here instead! That way you don't need to allocate each frame.

korangar/src/loaders/animation/mod.rs
kokosha: The hashmap is not using the entity id from the session, it is using the id linked with the sprite search in the Lua script. For now npc and monster is ok in cache, for player need to change later, i only passing the job id and sex for now and transform it in a number, later we can hash the parameters in other ways.

hasenbanck: Than please don't call it "entity_id". That's already a well established term!
If you know the ID's are unique (or make sure they are), then it's fine, but you can then also use an enum to make this better to use:

```
#[derive(Clone, Debug, Hash)]
enum AnimationKey {
    Player { job: u32, head: u32, is_female: bool },
    Monster(u32),
    Npc(u32),
}
```

korangar/src/graphics/passes/point_shadow/shader/entity_bindless.wgsl
kokosha: I changed this

korangar/src/graphics/passes/point_shadow/shader/entity_bindless.wgsl
kokosha: I documented the two part that I changed

korangar/src/graphics/passes/point_shadow/shader/entity_bindless.wgsl
kokosha: I want to try to shift using matrix, but my calculation abilities is limited...

korangar/src/graphics/passes/point_shadow/entity.rs
kokosha: I tested it works really well.

korangar/src/graphics/instruction.rs
kokosha: Yeah, i merged with depth_offset, it worked.

korangar/src/loaders/animation/mod.rs
kokosha: Changed to hashbrown and used the enum as key

korangar/src/loaders/animation/mod.rs
kokosha: Fixed

korangar/src/graphics/passes/point_shadow/shader/entity.wgsl
kokosha: Changed the information from every shader.

korangar/src/graphics/passes/point_shadow/shader/entity.wgsl
kokosha: Removed the extra parameters

korangar/src/graphics/passes/forward/entity.rs
kokosha: I was searching for Instruction in the repository it added a lunify =(

korangar/src/graphics/passes/forward/shader/entity_bindless.wgsl
kokosha: It the final alpha for transparency blending later.

korangar/src/graphics/instruction.rs
kokosha: This function is only for calculating the vertex, I don't know if we can convert the vertices of texture to texture_position and texture_size.

hasenbanck: If you can't simplify it for now it's fine. We can refine this part another time.

korangar/src/graphics/passes/point_shadow/shader/entity_bindless.wgsl
hasenbanck: It's fine for now. But could you please create a table in the documentation like we did with the other functions? That would help greatly if we want to debug this part later!

korangar/src/graphics/passes/directional_shadow/shader/entity_bindless.wgsl
```diff
@@ -47,26 +51,35 @@ fn vs_main(
...
+    output.position = pass_uniforms.view_projection * instance.world * new_vertex;
     output.texture_coordinates = instance.texture_position + vertex.texture_coordinates * instance.texture_size;
 
     if (instance.mirror != 0u) {
         output.texture_coordinates.x = 1.0 - output.texture_coordinates.x;
     }
 
-    output.depth_offset = vertex.depth_multiplier;
-    output.curvature = vertex.curvature_multiplier;
+    output.depth_offset = new_vertex.y/2.0 + instance.extra_depth_offset;
```
hasenbanck, line 65: This looks odd. Why are you now using the vertex x/y coordinates here?

korangar/src/graphics/passes/directional_shadow/shader/entity_bindless.wgsl
```diff
@@ -47,26 +51,35 @@ fn vs_main(
 ) -> VertexOutput {
     let instance = instance_data[instance_index];
     let vertex = vertex_data(vertex_index);
+    let new_vertex = instance.affine * vec4<f32>(vertex.position, 1.0);
```
hasenbanck, line 54: We calculate the world coordinate by using both the affine and the world matrix directly after each other. This means, that we can pre-multiply the affine and world matrices on the CPU side! (and don't need to send both to the GPU)

korangar/src/graphics/passes/directional_shadow/shader/entity_bindless.wgsl
kokosha: It is possible to calculate the depth and curvature using the x/y-coordinate. For depth_offset is y/2.0 and for curvature is x. In the original function of vertex it could be changed there also.

korangar/src/graphics/passes/directional_shadow/shader/entity_bindless.wgsl
kokosha: The problem is that for getting information about the depth and curvature is necessary to pass through the affine matrix.

korangar/src/graphics/passes/directional_shadow/shader/entity_bindless.wgsl
hasenbanck: Then you need to explain WHY this is possible in the code. Is it because it correlates to the X/Y value? I think the correct place for this is the vertex_data. Then you can also document it in the truth table.

korangar/src/graphics/passes/directional_shadow/shader/entity_bindless.wgsl
hasenbanck: Okay, I see. Btw: the variable "world_position" is never used, so you can remove it.
Also: please find a better names for "new_vertex" and "affine". I think the annimation is "morphing" the base entity part, correct? Maybe "morphed_vertex" and "morph" are better names?

korangar/src/graphics/passes/directional_shadow/shader/entity_bindless.wgsl
hasenbanck: Oh wait ... I think there is no need for chaneg the vertex_data() function at all and just us the old values (depth_multiplier and curvature_multiplier), since they have exactly the same value as you re-calculate here!

kokosha: You need to recalculate the depth and curvature unfortunately after the frame part transform. Another way is using a matrix multiplication with odd size to gather the result of depth and curvature, but it is also a problem because of the padding format in the shader side.

korangar/src/graphics/passes/directional_shadow/shader/entity_bindless.wgsl
kokosha: Morph is a difficult word, I changed to frame_part_vertex and frame_part_transform.

hasenbanck: The new name is good!

korangar/src/graphics/passes/directional_shadow/shader/entity_bindless.wgsl
hasenbanck: Then I write exactly what you just told me! The comment would be better this way:
```
// Because we have to transform the vertex of the frame part, we can't use the depth and curvature
// directly and are using the fast, that y/depth and x/curvature correlate to each other.
```

kokosha: Added the comment and also explained the offset.

hasenbanck: Great! LGTM!

PR #136, opened 2024-11-03
Lower camera rotation speed

eleriaqueen:
I suggest this small change as I feel Korangar's camera rotation is too fast but this may be because I'm used to the official client!
With `0.1` according to a few quick tests it feels more manageable / less twitchy and incidentally closer to the official client.
Feel free to test the PR and comment.


vE5li:
Feels good to me :+1:

hasenbanck:
Feels fine.

aleDsz:
LGTM! :shipit:

PR #137, opened 2024-11-04
Add support to `ChangeHair` network event

aleDsz:


kokosha:
can you make the following remap from id 0 - 12?
male = {2, 2, 1, 7, 5, 4, 3, 6, 8, 9, 10, 12, 11}
female = {2, 2, 4, 7, 1, 5, 3, 6, 12, 10, 9, 11, 8}

korangar/src/world/entity/mod.rs
```diff
@@ -251,12 +252,26 @@ fn get_entity_part_files(script_loader: &ScriptLoader, entity_type: EntityType,
...
     }
 
-    fn player_head_path(sex_sprite_path: &str, head_id: usize) -> String {
-        format!("ÀÎ°£Á·\\¸Ó¸®Åë\\{}\\{}_{}", sex_sprite_path, head_id, sex_sprite_path)
+    fn player_head_path(sex_sprite_path: &str, hair_id: usize) -> String {
+        format!("ÀÎ°£Á·\\¸Ó¸®Åë\\{}\\{}_{}", sex_sprite_path, hair_id, sex_sprite_path)
     }
 
+    let male_hair_ids = vec![2, 2, 1, 7, 5, 4, 3, 6, 8, 9, 10, 12, 11];
+    let female_hair_ids = vec![2, 2, 4, 7, 1, 5, 3, 6, 12, 10, 9, 11, 8];
```
vE5li, line 260: This should be a set of constants:

```rust
const MALE_HAIR_LOOKUP: &[usize] = &[2, 2, 1, 7, 5, 4, 3, 6, 8, 9, 10, 12, 11];
const FEMALE_HAIR_LOOKUP: &[usize] = &[2, 2, 4, 7, 1, 5, 3, 6, 12, 10, 9, 11, 8];
```

korangar/src/main.rs
```diff
@@ -1144,7 +1144,7 @@ impl Client {
                 NetworkEvent::UpdateEquippedPosition { index, equipped_position } => {
                     self.player_inventory.update_equipped_position(index, equipped_position);
                 }
-                NetworkEvent::ChangeJob(account_id, job_id) => {
+                NetworkEvent::ChangeSprite(account_id, sprite_type, value, value2) => {
```
vE5li, line 1147: This is not something I documented anywhere but `NetworkEvent`s should be a bit more abstract than the ragnrok online packets. I would like to keep the fact that changing the job and changing the hair sprite is sent via the same packet as an implemenation detail.

So in this case, I would like to keep the old `ChangeJob` event and _add_ a second event called `ChangeHead` (or something along these lines) to keep them separated.

korangar/src/world/entity/mod.rs
```diff
@@ -251,12 +252,26 @@ fn get_entity_part_files(script_loader: &ScriptLoader, entity_type: EntityType,
...
+
+    let hair_style_id = match (sex, hair_id) {
+        (Sex::Male, 0..12) => male_hair_ids[hair_id],
+        (Sex::Male, hair_id) => hair_id,
+        (Sex::Female, 0..12) => female_hair_ids[hair_id],
```
vE5li, line 265: Looks like the 12 here could be replaced by `MALE_HAIR_LOOKUP.len()` and `FEMALE_HAIR_LOOKUP.len()` respectively.

korangar_networking/src/event.rs
```diff
@@ -118,7 +118,7 @@ pub enum NetworkEvent {
         index: InventoryIndex,
         equipped_position: EquipPosition,
     },
-    ChangeJob(AccountId, u32),
+    ChangeSprite(AccountId, SpriteChangeType, u32, u32),
```
vE5li, line 121: When you revert the `ChangeSprite` and also when adding the second change packet, it would be amazing if you could convert from a tuple variant to a struct variant (e.g. `ChangeJob { account_id: AccountId, job_id: u32 }`). I've been changing all the variants to named fields one by one when working on them.

korangar/src/world/entity/mod.rs
```diff
@@ -67,6 +67,7 @@ pub enum EntityType {
 pub struct Common {
     pub entity_id: EntityId,
     pub job_id: usize,
+    pub hair_id: usize,
```
vE5li, line 70: Do entities besides the player also have a `hair_id`? I would assume it's only player characters, in wich case this should ot be in `Common`.

vE5li:
Looks good now, just one more thing :)

korangar/src/world/entity/mod.rs
```diff
@@ -767,6 +788,8 @@ impl Common {
 #[derive(PrototypeWindow)]
 pub struct Player {
     common: Common,
+
```
vE5li, line 791: Just noticed this change slipped in

PR #138, opened 2024-11-05
Implement CMAA2

hasenbanck:
It's shaper than FXAA. I found it best used with x2 or x4 MSAA, since they compliment each other well. I didn't port the MSAA mode, since it didn't increased to better overall image quality and was very performance hungry.

We could fine tune the overall parameters, but currently I think I hit a good sweet spot.

I had to return to the standard WebGPU limits, since we use compute shader more (shouldn't be a problem, since we don't support the downleveled backends, like DX11 or OpenGL anyhow).

I also hand to simplify the surface management by always rendering to texture and blitting as the last step to the surface. This enabled me to remove the limitation, that we only supported sRGB compatible surfaces. This isn't the case anymore and we now support normal UNORM and UNORM_SRGB now.

Please don't touch the shader code, since I explicitly left some quirks of the HLSL code in there, so that I could better compare bith versions, if such need should arise later.

### DIFF
--- a/korangar/src/graphics/capabilities.rs
+++ b/korangar/src/graphics/capabilities.rs
@@ -35,7 +35,7 @@ impl Capabilities {
             #[cfg(feature = "debug")]
             polygon_mode_line: false,
             required_features: Features::TEXTURE_ADAPTER_SPECIFIC_FORMAT_FEATURES,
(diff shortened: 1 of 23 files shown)


vE5li:
It hurts my soul a bit to merge the shaders like :smiling_face_with_tear:
It would be nice if you could do a cleanup PR when you're feeling confident enough in the shader

PR #139, opened 2024-11-06
Fix internal name of Thief and Royal Guard jobs

eleriaqueen:
Korangar would crash when trying to load assets for Thief job (id: 6), Royal Guard (id: 4066) and Royal Guard H (id: 4073)
With this fix Korangar can load their assets and display them.


aleDsz (approved):

vE5li:
I'm blindly trusting here :+1:

PR #140, opened 2024-11-06
Several fixes for entity rendering

kokosha:
The commit include the following fixes:

- Formatting the shader to ensure consistency across all entity shaders.
- Improving the text description of the algorithm used to generate animations for entities.
- The algorithm for generating entity positions has been updated. Instead of checking various offsets and applying small offset differences, it now involves identifying the entity's side extremities and creating a bounding box. The algorithm is simpler and easier to understand.
- When changing the player's appearance, pixels may sometimes shift by 1. To address this issue, the player is given a larger bounding box, ensuring the sprite is properly positioned within it.
- Implementing a temporary depth offset fix to prevent the lower part of the entity from clipping through walls.
- The rotation no longer causes the edges to be clipped and changed from being calculated inside the fragment shader to vertex shader.
- Fix the issue where the transparent parts of the entity are incorrectly displaying shadows.
- Resolve several bugs of entity calculation.


korangar/src/graphics/passes/directional_shadow/shader/entity_bindless.wgsl
```diff
@@ -71,20 +73,19 @@ fn vs_main(
...
 
 @fragment
 fn fs_main(input: VertexOutput) -> @builtin(frag_depth) f32 {
-    // Apply the rotation from action
-    let sin_factor = sin(input.angle);
-    let cos_factor = cos(input.angle);
-    let rotate = vec2(input.texture_coordinates.x - 0.5, input.texture_coordinates.y - 0.5) * mat2x2(cos_factor, sin_factor, -sin_factor, cos_factor);
-    let texture_coordinates = vec2(clamp(rotate.x + 0.5, 0.0, 1.0), clamp(rotate.y + 0.5, 0.0, 1.0));
+    let action_rotation_matrix = instance_data[input.instance_index].action_rotation_matrix;
+    let texture_coordinates = action_rotation_transform(input.texture_coordinates, action_rotation_matrix);
```
hasenbanck, line 85: You are using the interpolated value from the fragment shader directly here to apply the action_rotation_transform(). This means you should be able to move this function to the vertex shader. At least in my local test this works perfectly fine!

korangar/src/graphics/passes/forward/entity.rs
```diff
@@ -28,16 +28,17 @@ const INITIAL_INSTRUCTION_SIZE: usize = 256;
...
-    color: [f32; 4],
     extra_depth_offset: f32,
     depth_offset: f32,
-    angle: f32,
     curvature: f32,
     mirror: u32,
     texture_index: i32,
-    padding: [u32; 2],
+    padding: u32,
+    padding_2: [u32; 2],
```
hasenbanck, line 41: No need to use padding_2 here. Just make "padding" a [u32; 2].

korangar/src/loaders/animation/mod.rs
```diff
@@ -297,9 +319,8 @@ impl AnimationLoader {
 /// inside the frame bounding rectangle with vertices [-1, 0], [-1, 2], [1, 0],
 /// [1, 2].
 fn convert_coordinates(coordinates: Vector2<i32>, size: Vector2<i32>) -> Vector2<f32> {
-    const EPSILON: f32 = 0.0001;
-    let x = (coordinates.x as f32 / size.x as f32 - 0.5) * 2.0 + EPSILON;
-    let y = 2.0 - (coordinates.y as f32 / size.y as f32) * 2.0 + EPSILON;
+    let x = (coordinates.x as f32 / size.x as f32 - 0.5) * 2.0;
```
hasenbanck, line 322: The old and new code don't properly guard against division by zero (if either dimension of "size" is 0).

hasenbanck (changes requested):
Some comments mainly about the shaders and correctness of the code.

korangar/src/graphics/passes/forward/entity.rs
kokosha: Was having some issue with depth offset and think that the shader was wrong, because of padding, but I was wrong

korangar/src/loaders/animation/mod.rs
kokosha: Panic when the size is 0, the construction doesn't allow, but for debugging purposes would be great!

korangar/src/loaders/animation/mod.rs
```diff
@@ -176,95 +183,104 @@ impl AnimationLoader {
...
+                    let rotation_matrix = Matrix4::from_angle_z(Rad(-frame_part.angle));
 
                     let new_center = Vector2::new(0.0, (texture_top_left.y - texture_bottom_left.y) / 2.0);
                     let texture_center = (texture_top_left + texture_bottom_right) / 2.0;
                     let translation_matrix = Matrix4::from_translation(texture_center.extend(1.0) - new_center.extend(1.0));
 
-                    frame_part.affine_matrix = translation_matrix * scale_matrix;
+                    frame_part.affine_matrix = translation_matrix
+                        * scale_matrix
+                        * Matrix4::from_translation(Vector3::<f32>::new(0.0, 1.0, 0.0))
```
hasenbanck, line 279: Can you please document why we need to do the two translations? Are they just random stuff that gravity applied?

korangar/src/loaders/animation/mod.rs
```diff
@@ -297,9 +313,11 @@ impl AnimationLoader {
 /// inside the frame bounding rectangle with vertices [-1, 0], [-1, 2], [1, 0],
 /// [1, 2].
 fn convert_coordinates(coordinates: Vector2<i32>, size: Vector2<i32>) -> Vector2<f32> {
-    const EPSILON: f32 = 0.0001;
-    let x = (coordinates.x as f32 / size.x as f32 - 0.5) * 2.0 + EPSILON;
-    let y = 2.0 - (coordinates.y as f32 / size.y as f32) * 2.0 + EPSILON;
+    if size.x == 0 || size.y == 0 {
+        panic!("convert_coordinates - The size is not possible to be 0 as how it was created");
```
hasenbanck, line 317: You can just do a "assert!(size.x != 0 && size.y != 0);" here I guess.

korangar/src/loaders/animation/mod.rs
kokosha: Ok, fixed the documentation of the translation.
May I change the order of translation to make it more safe?

kokosha: I will change a multiplication by a division.

hasenbanck: Sure! Safer is always better!

korangar/src/graphics/passes/point_shadow/entity.rs
```diff
@@ -28,11 +28,11 @@ struct InstanceData {
     texture_size: [f32; 2],
     extra_depth_offset: f32,
     depth_offset: f32,
-    angle: f32,
     curvature: f32,
     mirror: u32,
     texture_index: i32,
-    padding: [u32; 2],
+    padding: u32,
+    padding_2: [u32; 2],
```
hasenbanck, line 35: Can you please simplify the padding here too?

kokosha: Opsss, forget to change... Good catch!

korangar/src/graphics/passes/directional_shadow/shader/entity.wgsl
```diff
@@ -60,29 +60,30 @@ fn vs_main(
         output.texture_coordinates.x = 1.0 - output.texture_coordinates.x;
     }
 
+    let MAX_X = 500.0;
```
hasenbanck, line 63: Why do we use 500 here?

korangar/src/graphics/passes/forward/shader/entity.wgsl
```diff
@@ -107,30 +105,30 @@ fn vs_main(
...
-    let cos_factor = cos(input.angle);
-    let rotate = vec2(input.texture_coordinates.x - 0.5, input.texture_coordinates.y - 0.5) * mat2x2(cos_factor, sin_factor, -sin_factor, cos_factor);
-    let texture_coordinates = vec2(clamp(rotate.x + 0.5, 0.0, 1.0), clamp(rotate.y + 0.5, 0.0, 1.0));
-
-    let diffuse_color = textureSample(texture, texture_sampler, texture_coordinates);
-    let alpha_channel = textureSample(texture, nearest_sampler, texture_coordinates).a;
+    let diffuse_color = textureSample(texture, texture_sampler, input.texture_coordinates);
+    let alpha_channel = textureSample(texture, nearest_sampler, input.texture_coordinates).a;
 
     if (alpha_channel == 0.0) {
```
hasenbanck, line 133: Don't we need to also discard here if the input.color alpha is not 1.0?

hasenbanck:
I had a quick test and the position of the entity sprites look very good! 
But maybe we need to adjust the steepness of the depth ofset a little more? I personaly have a simple test:

If the indicator shows, that the entity is clearly "in front of" 3D geometry, the entity sprite should not clip into it:
Screenshot 2024-11-17 084917

Right not the tip of the head sprite seems to clip into 3D geometry in such cases:

Screenshot 2024-11-17 084856

korangar/src/graphics/passes/directional_shadow/shader/entity.wgsl
kokosha: It the sprite max size will change the naming

korangar/src/graphics/passes/forward/shader/entity.wgsl
kokosha: Yes, I will change there.

korangar/src/graphics/passes/directional_shadow/shader/entity.wgsl
kokosha: I am using the wrong value to calculate the proportion, I will fix it.

kokosha:
> I had a quick test and the position of the entity sprites look very good! But maybe we need to adjust the steepness of the depth ofset a little more? I personaly have a simple test:
> 
> If the indicator shows, that the entity is clearly "in front of" 3D geometry, the entity sprite should not clip into it: Screenshot 2024-11-17 084917
> 
> Right not the tip of the head sprite seems to clip into 3D geometry in such cases:
> Screenshot 2024-11-17 084856

I corrected the original depth offset method to the right frame size, but there is another depth offset error caused by clipping in the bottom part of pillar, for correcting this would need to change the method of depth offset.

hasenbanck:
Thead now doesn't clip the above test. Looks fine to me now.

vE5li:
The PR looks good but there are some things I would ask for a pass on

### Formattting
Some equasions in the shaders are formatted like this
```wgsl
let proportion_x = instance.frame_size.x/SPRITE_MAX_SIZE_X;
let proportion_y = instance.frame_size.y/SPRITE_MAX_SIZE_Y;
```
We always add spaces around binary operators, e.g. `instance.frame_size.x / SPRITE_MAX_SIZE_X;`

### Grammar
A lot of the comments don't use correct grammar, I would ask you to please double check. I can recommend ChatGPT or other LLMs for getting feedback on texts.

### Newlines
A lot of the files you worked on don't end on a newline, which is why GitHub is showing this symbol
no-newline

Basically every editor should allow you to enable adding this newline automatically, so you don't even have to think about it in the future.

hasenbanck:
I took a final check and saw following issues left to do:

1. Please have a look at the github file review. There are still some files left where you removed the empty line at the end.
2. Can you please port the changes you did from the directional shadow entity shader to the point shadow entity shader? They look very different.
3. Final rebase

hasenbanck (approved):

PR #142, opened 2024-11-08
Fix RSW with object that node name doesn't have \0

kokosha:
The map iz_d04_i have objects that have node name that doesn't have \0, the entire 80 bytes from the string have nothing but random bytes.


vE5li:
I don't really understand this PR. The `String` that `pub _node_name` was using before also doesn't allow to have any `0` bytes, what does this actually change?

kokosha:
Hello! The game crash in the map iz_d04_i, because the object inside RSW file have a pub _node_name with 80 bytes and none of the 80 bytes is \0. 

When reading as a string, you need to have a delimiter \0 to read. In the current code, it will reach the end of byte stream without find a \0 and crash as the byte stream is too short.

For testing purpose, you could add a warp to the map.

vE5li:
I guess what confused me is that `iz_d04_i` loads just fine for me, but maybe we have different versions.

Given the scenario you described, I would rather update the `String` parsing instead to not panic when it runs out of bytes. So basically changing the impementation to

```rust
        while let Ok(byte) = byte_stream.byte::<Self>() {
            match byte {
                0 => break,
                byte => value.push(byte as char),
            }
        }
```

Do you mind adjusting the PR and getting it ready?

PR #146, opened 2024-11-14
RSM bytestream compatible to 2.3

kokosha:
Update the RSM bytestream to be compatible with version 2.2 and 2.3.

### DIFF
--- a/korangar/src/loaders/model/mod.rs
+++ b/korangar/src/loaders/model/mod.rs
@@ -66,7 +66,7 @@ impl ModelLoader {
     fn make_vertices(node: &NodeData, main_matrix: &Matrix4<f32>, reverse_order: bool) -> Vec<NativeModelVertex> {
         let mut native_vertices = Vec::new();
 
-        let array: [f32; 3] = node.scale.into();
+        let array: [f32; 3] = node.scale.unwrap().into();
         let reverse_node_order = array.into_iter().fold(1.0, |a, b| a * b).is_sign_negative();
 
         if reverse_node_order {
@@ -115,10 +115,10 @@ impl ModelLoader {
     }
 
     fn calculate_matrices(node: &NodeData, parent_matrix: &Matrix4<f32>) -> (Matrix4<f32>, Matrix4<f32>, Matrix4<f32>) {
-        let main = Matrix4::from_translation(node.translation1) * Matrix4::from(node.offset_matrix);
-
-        let scale_matrix = Matrix4::from_nonuniform_scale(node.scale.x, node.scale.y, node.scale.z);
-        let rotation_matrix = Matrix4::from_axis_angle(node.rotation_axis, Rad(node.rotation_angle));
+        let main = Matrix4::from_translation(node.translation1.unwrap()) * Matrix4::from(node.offset_matrix);
+        let scale = node.scale.unwrap();
+        let scale_matrix = Matrix4::from_nonuniform_scale(scale.x, scale.y, scale.z);
+        let rotation_matrix = Matrix4::from_axis_angle(node.rotation_axis.unwrap(), Rad(node.rotation_angle.unwrap()));
         let translation_matrix = Matrix4::from_translation(node.translation2);
 
         let transform = match node.rotation_keyframe_count > 0 {
@@ -247,6 +247,9 @@ impl ModelLoader {
         };
 
         // TODO: Temporary check until we support more versions.
+        // TODO: The model operation to scale keyframe is not implemented yet.
+        // TODO: The model operation to translate keyframe is not implemented yet.
+        // TODO: The model operation to modify texture keyframe is not implemented yet.
         let version: InternalVersion = model_data.version.into();
         if version.equals_or_above(2, 2) {
             #[cfg(feature = "debug")]
@@ -265,7 +268,7 @@ impl ModelLoader {
             .collect();
         let texture_mapping: Vec<i32> = (0..model_data.texture_names.len()).map(|index| index as i32).collect();
 
-        let root_node_name = &model_data.root_node_name;
+        let root_node_name = &model_data.root_node_name.clone().unwrap();
         let root_node = model_data
             .nodes
             .iter()
--- a/ragnarok_formats/src/model.rs
+++ b/ragnarok_formats/src/model.rs
@@ -6,7 +6,7 @@ use ragnarok_bytes::{
 use crate::signature::Signature;
 use crate::version::{InternalVersion, MajorFirst, Version};
 
-/// A string that can either have a fixed lenght or be length prefixed, based on
+/// A string that can either have a fixed length or be length prefixed, based on
 /// the file format version.
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct ModelString<const LENGTH: usize> {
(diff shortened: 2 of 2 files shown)


korangar/src/loaders/model/mod.rs
```diff
@@ -116,10 +116,10 @@ impl ModelLoader {
 
     fn calculate_matrices(node: &NodeData, parent_matrix: &Matrix4<f32>) -> (Matrix4<f32>, Matrix4<f32>, Matrix4<f32>) {
         let main = Matrix4::from_translation(node.translation1) * Matrix4::from(node.offset_matrix);
-
-        let scale_matrix = Matrix4::from_nonuniform_scale(node.scale.x, node.scale.y, node.scale.z);
-        let rotation_matrix = Matrix4::from_axis_angle(node.rotation_axis, Rad(node.rotation_angle));
-        let translation_matrix = Matrix4::from_translation(node.translation2);
+        let scale_size = node.scale.unwrap();
```
vE5li, line 119: `scale_size` doesn't really make a lot of sense, just `scale` would be better

ragnarok_formats/src/model.rs
```diff
@@ -57,21 +57,64 @@ where
...
+#[derive(Debug, ByteConvertable)]
+#[cfg_attr(feature = "interface", derive(korangar_interface::elements::PrototypeElement))]
+pub struct TextureKeyframeData {
+    // List of operation types - https://rathena.org/board/topic/127587-rsm2-file-format/
+    // 0: Texture translation on the X axis. The texture is tiled.
+    // 1: Texture translation on the Y axis. The texture is tiled.
+    // 2: Texture multiplication on the X axis. The texture is tiled.
+    // 3: Texture multiplication on the Y axis. The texture is tiled
+    // 4: Texture rotation around (0, 0). The texture is not tiled.
+    pub operation_type: u32,
```
vE5li, line 99: Could we make this be an enum?

ragnarok_formats/src/model.rs
```diff
@@ -115,15 +168,28 @@ pub struct NodeData {
...
-    #[repeating_option(position_keyframe_count)]
-    pub position_keyframes: Vec<PositionKeyframeData>,
+    pub scale_keyframe_count: Option<u32>,
+    #[repeating_option(scale_keyframe_count)]
+    pub scale_keyframes: Vec<ScaleKeyframeData>,
     #[new_derive]
     pub rotation_keyframe_count: u32,
     #[repeating(rotation_keyframe_count)]
     pub rotation_keyframes: Vec<RotationKeyframeData>,
+    // TODO: Translation Keyframe not implemented yet
```
vE5li, line 181: What do you mean by not yet implemented? The parsing or using the data?

vE5li (commented):
Looks good! Just some small comments/questions

ragnarok_formats/src/model.rs
kokosha: The manipulation of the model

ragnarok_formats/src/model.rs
kokosha: I will change to use a enum

ragnarok_formats/src/model.rs
vE5li: The we should either remove the comments or put it somewhere else. Something like `TODO: Use translations keyframes` in the `load` method of the `ModelLoader` is less likely to become a stale comment.

ragnarok_formats/src/model.rs
```diff
@@ -57,21 +57,74 @@ where
...
 pub struct RotationKeyframeData {
     pub frame: u32,
     pub quaternions: Quaternion<f32>,
 }
 
+#[derive(Debug, ByteConvertable)]
+#[cfg_attr(feature = "interface", derive(korangar_interface::elements::PrototypeElement))]
+pub struct TranslationKeyframeData {
+    pub frame: u32,
+    pub translate: Vector3<f32>,
```
vE5li, line 76: Given that the struct is called `TranslationKeyframeData`, I think this should be `translation`

korangar/src/loaders/model/mod.rs
```diff
@@ -247,6 +247,9 @@ impl ModelLoader {
         };
 
         // TODO: Temporary check until we support more versions.
+        // TODO: The model operation to scale keyframe not implemented yet
+        // TODO: The model operation to translate keyframe not implemented yet
+        // TODO: The model operation to modify texture keyframe not implemented yet
```
vE5li, line 252: Please make sure that the comments use correct grammar (and end them on a `.` please)

ragnarok_formats/src/model.rs
```diff
@@ -57,21 +57,74 @@ where
...
+// 4: Texture rotation around (0, 0). The texture is not tiled.
+#[derive(Debug, ByteConvertable)]
+#[cfg_attr(feature = "interface", derive(korangar_interface::elements::PrototypeElement))]
+pub enum TextureOperation {
+    TranslationX,
+    TranslationY,
+    ScaleX,
+    ScaleY,
+    Rotation,
+}
```
vE5li, line 104: For documentation style comments like this you should really use Rusts doc comments (`///`). That will make these comments show up in peoples IDEs as well as documentation generated by `cargo doc`.

You can attach doc comments to almost anything, for example the varians of an enum. So I would instead write the coments like this:

```rust
/// List of texture operation types.
/// See: https://rathena.org/board/topic/127587-rsm2-file-format/
#[derive(Debug, ByteConvertable)]
#[cfg_attr(feature = "interface", derive(korangar_interface::elements::PrototypeElement))]
pub enum TextureOperation {
    /// Texture translation on the X axis. The texture is tiled.
    TranslationX,
    /// Texture translation on the Y axis. The texture is tiled.
    TranslationY,
    /// Texture multiplication on the X axis. The texture is tiled.
    ScaleX,
    /// Texture multiplication on the Y axis. The texture is tiled
    ScaleY,
    /// Texture rotation around (0, 0). The texture is not tiled.
    Rotation,
}
```

ragnarok_formats/src/model.rs
```diff
@@ -57,27 +57,84 @@ where
...
+    #[version_equals_or_above(2, 2)]
+    pub length: Option<u32>,
     pub vertex_position_indices: [u16; 3],
     pub texture_coordinate_indices: [u16; 3],
     pub texture_index: u16,
     pub padding: u16,
     pub two_sided: i32,
     pub smooth_group: i32,
+    #[version_equals_or_above(2, 2)]
+    #[repeating_expr(usize::max(0, (length.unwrap() as usize - 24)/4))]
```
vE5li, line 136:
```suggestion
    #[repeating_expr(0.max((length.unwrap() as usize - 24) / 4))]
```

ragnarok_formats/src/model.rs
```diff
@@ -92,17 +149,32 @@ pub struct TextureCoordinateData {
...
-    pub scale: Vector3<f32>,
+    #[version_smaller(2, 2)]
+    #[new_default]
+    pub rotation_angle: Option<f32>,
+    #[version_smaller(2, 2)]
+    #[new_default]
+    pub rotation_axis: Option<Vector3<f32>>,
+    #[version_smaller(2, 2)]
+    #[new_default]
+    pub scale: Option<Vector3<f32>>,
```
vE5li, line 177: Why are `translation1`, `rotation_angle`, `rotation_axis` and `scale` marked with `new_default`?

`new_default` is meant for fields that the user should never care about setting when creating a new instance of the type (like the signature, unknown bytes, patting, etc.).

EDIT: Actually, same for `frames_per_second`

vE5li:
After you applied those changes please squash the commit history so I can merge :)

ragnarok_formats/src/model.rs
```diff
@@ -57,27 +57,84 @@ where
...
+    #[version_equals_or_above(2, 2)]
+    pub length: Option<u32>,
     pub vertex_position_indices: [u16; 3],
     pub texture_coordinate_indices: [u16; 3],
     pub texture_index: u16,
     pub padding: u16,
     pub two_sided: i32,
     pub smooth_group: i32,
+    #[version_equals_or_above(2, 2)]
+    #[repeating_expr(0.max((length.unwrap() as usize - 24) / 4))]
```
vE5li, line 136: My bad, I just realized the `0.max` is also redundant. The type is already `usize` so it can't be negative

PR #147, opened 2024-11-16
Unifies the fallback loading

hasenbanck:
Moved the fallback handling in the specific loader of the resource. The specific loader can then decide, if it's a hard error or a fallback is needed.

This also adds a fallback in case an image format is not supported yet (for example BINK video files).

### DIFF
--- a/korangar/src/loaders/mod.rs
+++ b/korangar/src/loaders/mod.rs
@@ -24,3 +24,10 @@ pub use self::script::{ResourceMetadata, ScriptLoader};
 pub use self::server::{load_client_info, ClientInfo, ServiceId};
 pub use self::sprite::*;
 pub use self::texture::{TextureAtlasFactory, TextureLoader};
+
+pub const FALLBACK_PNG_FILE: &str = "missing.png";
+pub const FALLBACK_BMP_FILE: &str = "missing.bmp";
+pub const FALLBACK_TGA_FILE: &str = "missing.tga";
+pub const FALLBACK_MODEL_FILE: &str = "missing.rsm";
+pub const FALLBACK_SPRITE_FILE: &str = "npc\\missing.spr";
+pub const FALLBACK_ACTIONS_FILE: &str = "npc\\missing.act";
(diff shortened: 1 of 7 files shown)


korangar/src/loaders/mod.rs
```diff
@@ -24,3 +24,10 @@ pub use self::script::{ResourceMetadata, ScriptLoader};
 pub use self::server::{load_client_info, ClientInfo, ServiceId};
 pub use self::sprite::*;
 pub use self::texture::{TextureAtlasFactory, TextureLoader};
+
+pub const FALLBACK_PNG_FILE: &str = "missing.png";
+pub const FALLBACK_BMP_FILE: &str = "missing.bmp";
```
vE5li, line 29: `FALLBACK_BMP_FILE` and `FALLBACK_TGA_FILE` are still there but not used anymore. They do actually look different so I would either unify them into a `FALLBACK_TEXTURE_FILE` or properly check which one to return on fallback.

hasenbanck: We still use them here: 

https://github.com/vE5li/korangar/blob/e98be9b802a7c5a8b9daae05827396f4e68e93d1/korangar/src/loaders/texture/mod.rs#L100

Or do you mean that you want to differentiate when we don't find the image data here?

https://github.com/vE5li/korangar/blob/e98be9b802a7c5a8b9daae05827396f4e68e93d1/korangar/src/loaders/texture/mod.rs#L86

korangar_util/src/loader.rs
```diff
@@ -10,6 +9,12 @@ impl FileNotFoundError {
     }
 }
 
+impl std::fmt::Debug for FileNotFoundError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "Can't find file: {}", self.0)
```
vE5li, line 14: Actually super small nit: I am not super consisent with this myself but it is best practice to start error messages (and panic messages) with a lowercase character (and end without a trailing dot)

hasenbanck: Fixed

PR #148, opened 2024-11-17
Simplify the cache and use it everywhere

hasenbanck:
The ResourceCache was not straightforward to use, so I simplified it. This PR uses the SimpleCache for all caches that we have. Sine the cache statistics are exposed, we could include a debug view that shows the count and size of each cache. But I think that should be done, once the new UI refactor is done.


korangar_util/src/container/simple_cache.rs
```diff
@@ -0,0 +1,264 @@
...
+        Q: Hash + Eq + ?Sized,
+    {
+        if let Some(cache_key) = self.lookup.remove(key) {
+            self.cache.remove(cache_key);
+            let value = self.values.remove(cache_key);
+            self.update_statistics();
+            value
+        } else {
+            None
+        }
```
vE5li, line 131: I think this could be done more elegantly with `and_then`

korangar_util/src/container/simple_cache.rs
```diff
@@ -0,0 +1,264 @@
...
+    where
+        K: Borrow<Q>,
+        Q: Hash + Eq + ?Sized,
+    {
+        if let Some(&cache_key) = self.lookup.get(key) {
+            self.cache.touch(cache_key);
+            self.values.get(cache_key)
+        } else {
+            None
+        }
```
vE5li, line 114: Same here

vE5li (commented):
Looks good!

korangar_util/src/container/simple_cache.rs
hasenbanck: Fixed

korangar_util/src/container/simple_cache.rs
hasenbanck: Fixed

PR #149, opened 2024-11-19
Remove the white border in entity

kokosha:
The goal is to remove the white border in entity.
The alpha blending is bleeding in the border of the texture, because is sampling the transparent part of picture in weighted average, resulting in a white pixel. To correct the problem, the transparent border that is used in weighted average is recolored with color close to the opaque border.
with_border_recoloring
without_border_recoloring

The depth buffer depends on the order of drawing, if you draw a transparent pixel, you can't draw a opaque pixel before the transparent pixel, to fix the problem, we sort by the value of the distance between the camera and the entity and render from farthest to nearest distance to camera. 
depth_buffer
depth_buffer_transparent

A small fix is made in the picker, after sorting the entities, the initial entries is not the current player anymore.

### DIFF
--- a/korangar/src/graphics/passes/picker/entity.rs
+++ b/korangar/src/graphics/passes/picker/entity.rs
@@ -214,6 +214,7 @@ impl Drawer<{ BindGroupCount::One }, { ColorAttachmentCount::One }, { DepthAttac
 
 impl Prepare for PickerEntityDrawer {
     fn prepare(&mut self, device: &Device, instructions: &RenderInstruction) {
+        // TODO: Pass the player size directly
         let player_size = 2;
         self.draw_count = instructions.entities.len().saturating_sub(player_size);
 
@@ -232,7 +233,7 @@ impl Prepare for PickerEntityDrawer {
             // We skip the first entity, because we don't want the player entity to show up
             // in the picker buffer.
             // TODO: Remove the player entity correctly.
-            for instruction in instructions.entities.iter().skip(player_size) {
+            for instruction in instructions.entities.iter().filter(|instruction| instruction.pickerable == true) {
                 let picker_target = PickerTarget::Entity(instruction.entity_id);
                 let (identifier_high, identifier_low) = picker_target.into();
 
@@ -267,7 +268,7 @@ impl Prepare for PickerEntityDrawer {
             // We skip the first entity, because we don't want the player entity to show up
             // in the picker buffer.
             // TODO: Remove the player entity correctly.
-            for instruction in instructions.entities.iter().skip(player_size) {
+            for instruction in instructions.entities.iter().filter(|instruction| instruction.pickerable == true) {
                 let picker_target = PickerTarget::Entity(instruction.entity_id);
                 let (identifier_high, identifier_low) = picker_target.into();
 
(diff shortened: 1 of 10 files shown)


korangar/src/graphics/passes/picker/entity.rs
```diff
@@ -229,10 +230,10 @@ impl Prepare for PickerEntityDrawer {
 
             let mut texture_views = Vec::with_capacity_in(self.draw_count, &self.bump);
 
-            // We skip the first entity, because we don't want the player entity to show up
-            // in the picker buffer.
-            // TODO: Remove the player entity correctly.
-            for instruction in instructions.entities.iter().skip(player_size) {
+            for instruction in instructions.entities.iter() {
```
hasenbanck, line 233: I think we can use `iter().filter()` here

korangar/src/graphics/passes/picker/entity.rs
```diff
@@ -264,10 +265,10 @@ impl Prepare for PickerEntityDrawer {
             self.instance_data_buffer.reserve(device, self.instance_data.len());
             self.bind_group = Self::create_bind_group_bindless(device, &self.bind_group_layout, &self.instance_data_buffer, &texture_views)
         } else {
-            // We skip the first entity, because we don't want the player entity to show up
-            // in the picker buffer.
-            // TODO: Remove the player entity correctly.
-            for instruction in instructions.entities.iter().skip(player_size) {
+            for instruction in instructions.entities.iter() {
```
hasenbanck, line 268: I think we can use iter().filter() here too.

korangar/src/world/map/mod.rs
```diff
@@ -265,10 +265,27 @@ impl Map {
...
     #[cfg_attr(feature = "debug", korangar_debug::profile)]
     pub fn render_entities(&self, instructions: &mut Vec<EntityInstruction>, entities: &[Entity], camera: &dyn Camera, include_self: bool) {
-        entities
-            .iter()
-            .skip(!include_self as usize)
-            .for_each(|entity| entity.render(instructions, camera));
+        fn get_distance(camera: &dyn Camera, entity: &Entity) -> f32 {
+            camera.distance_to(entity.get_position())
+        }
+        let mut entity_order: Vec<usize> = (0..entities.len()).collect();
```
hasenbanck, line 271: We should try to not allocate every frame inside the hot loop.

korangar/src/world/map/mod.rs
```diff
@@ -265,10 +265,27 @@ impl Map {
...
+
+        entity_order.sort_by(|index_a, index_b| {
+            let dist_a = get_distance(camera, &entities[*index_a]);
+            let dist_b = get_distance(camera, &entities[*index_b]);
+            dist_b.partial_cmp(&dist_a).unwrap()
+        });
+
+        entity_order.iter().for_each(|&index| {
+            let is_self = index == 0;
```
hasenbanck, line 280: "self" has a pretty well defined meaning in Rust. We should better use "is_player_entity" or somthing similar.

kokosha:
The new way is changing the alpha from (0.0;1.0] into 1.0, when the entity is opaque. When the entity is transparent, we need to sort to treat the ordering, but even sorting won't solve it all, as the first transparent pixel will overwrite the depth buffer, you will need new techniques such as depth peeling or other order-independent transparency method and is process intensive to fix transparency. Another problem is the way that the entities is structured in the moment doesn't allow easy sorting without affecting hot loop.

kokosha:
[156](https://github.com/vE5li/korangar/pull/156),  [157](https://github.com/vE5li/korangar/pull/157), fixed the problem

PR #156, opened 2024-11-26
Implement and use premultiplied alpha

hasenbanck:
Since this fixes one of the issues with the white border around the entities, we should switch to pre-multiplied alpha throughout the render pipeline.

I did my best to catch all places where we had to change the logic when handling alpha values.

I moved the premultiply_alpha() function into the util crate, so that it gets properly optimized when compiling debug builds.

grafik

### DIFF
--- a/korangar/src/graphics/passes/postprocessing/blitter.rs
+++ b/korangar/src/graphics/passes/postprocessing/blitter.rs
@@ -165,7 +165,11 @@ impl PostProcessingBlitterDrawer {
                 },
                 targets: &[Some(ColorTargetState {
                     format: color_texture_format,
-                    blend: if alpha_blending { Some(BlendState::ALPHA_BLENDING) } else { None },
+                    blend: if alpha_blending {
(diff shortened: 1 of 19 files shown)


korangar/src/graphics/passes/postprocessing/blitter.rs
```diff
@@ -165,7 +165,11 @@ impl PostProcessingBlitterDrawer {
                 },
                 targets: &[Some(ColorTargetState {
                     format: color_texture_format,
-                    blend: if alpha_blending { Some(BlendState::ALPHA_BLENDING) } else { None },
+                    blend: if alpha_blending {
+                        Some(BlendState::PREMULTIPLIED_ALPHA_BLENDING)
+                    } else {
+                        None
+                    },
```
vE5li, line 172:
```suggestion
                    blend: alpha_blending.then_some(BlendState::PREMULTIPLIED_ALPHA_BLENDING),
```

vE5li (approved):
Very small comment

PR #160, opened 2024-11-28
Use core::array::try_from_fn for deserializing arrays

kokosha:
The goal is to use core::array::try_from_fn for deserializing arrays.


vE5li:
Thank you for the PR but I already closed the issue as resolved

PR #161, opened 2024-11-28
Fix some small typos

kokosha:


korangar/src/loaders/script/mod.rs
```diff
@@ -131,7 +131,7 @@ end
         let globals = self.state.globals();
         let lookup_name = match is_identified {
             true => "identifiedResourceName",
-            false => "undidentifiedResourceName",
+            false => "unidentifiedResourceName",
```
vE5li, line 134: Did you verify that the current name is wrong? Because a lot of names in Ragnarok Online are incorrect/misspelled

vE5li (commented):
One question, also the commit should be something like `Fix typos`, and not `Find`

korangar/src/loaders/script/mod.rs
kokosha: The game crashed and I fixed it.

PR #162, opened 2024-11-28
Explicitly encode if an entity should be included in the picker buffer

kokosha:
After sorting the entities from back to front, the logic from picker ignoring first two elements need to be changed.

### DIFF
--- a/korangar/src/graphics/instruction.rs
+++ b/korangar/src/graphics/instruction.rs
@@ -185,6 +185,7 @@ pub struct EntityInstruction {
     pub color: Color,
     pub mirror: bool,
     pub entity_id: EntityId,
+    pub add_to_picker: bool,
     pub texture: Arc<Texture>,
     pub distance: f32,
 }
--- a/korangar/src/graphics/passes/picker/entity.rs
+++ b/korangar/src/graphics/passes/picker/entity.rs
@@ -199,90 +199,102 @@ impl Drawer<{ BindGroupCount::One }, { ColorAttachmentCount::One }, { DepthAttac
             let mut current_texture_id = self.solid_pixel_texture.get_id();
             pass.set_bind_group(2, self.solid_pixel_texture.get_bind_group(), &[]);
 
-            for (index, instruction) in draw_data[1..self.draw_count.saturating_add(1)].iter().enumerate() {
-                if instruction.texture.get_id() != current_texture_id {
-                    current_texture_id = instruction.texture.get_id();
-                    pass.set_bind_group(2, instruction.texture.get_bind_group(), &[]);
-                }
-                let index = index as u32;
+            draw_data
+                .iter()
+                .filter(|instruction| instruction.add_to_picker)
+                .enumerate()
+                .for_each(|(index, instruction)| {
+                    if instruction.texture.get_id() != current_texture_id {
+                        current_texture_id = instruction.texture.get_id();
+                        pass.set_bind_group(2, instruction.texture.get_bind_group(), &[]);
+                    }
+                    let index = index as u32;
 
-                pass.draw(0..6, index..index + 1);
-            }
+                    pass.draw(0..6, index..index + 1);
+                });
         }
     }
 }
 
 impl Prepare for PickerEntityDrawer {
     fn prepare(&mut self, device: &Device, instructions: &RenderInstruction) {
-        let player_size = 2;
-        self.draw_count = instructions.entities.len().saturating_sub(player_size);
-
-        if self.draw_count == 0 {
-            return;
-        }
-
         self.instance_data.clear();
+        self.draw_count = 0;
 
         if self.bindless_support {
             self.bump.reset();
             self.lookup.clear();
 
-            let mut texture_views = Vec::with_capacity_in(self.draw_count, &self.bump);
+            let mut texture_views = Vec::with_capacity_in(instructions.entities.len(), &self.bump);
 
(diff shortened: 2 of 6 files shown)


korangar/src/graphics/passes/picker/entity.rs
```diff
@@ -214,6 +214,7 @@ impl Drawer<{ BindGroupCount::One }, { ColorAttachmentCount::One }, { DepthAttac
 
 impl Prepare for PickerEntityDrawer {
     fn prepare(&mut self, device: &Device, instructions: &RenderInstruction) {
+        // TODO: Pass the player size directly instead of hardcoding it here.
```
hasenbanck, line 217:
```rust
fn prepare(&mut self, device: &Device, instructions: &RenderInstruction) {
        self.instance_data.clear();
        self.draw_count = 0;

        if self.bindless_support {
            self.bump.reset();
            self.lookup.clear();

            let mut texture_views = Vec::with_capacity_in(self.draw_count, &self.bump);

            for instruction in instructions.entities.iter().filter(|instruction| instruction.pickerable) {
                self.draw_count += 1;
```

Let's simplify this like this.

korangar/src/graphics/passes/picker/entity.rs
```diff
@@ -264,10 +262,7 @@ impl Prepare for PickerEntityDrawer {
             self.instance_data_buffer.reserve(device, self.instance_data.len());
             self.bind_group = Self::create_bind_group_bindless(device, &self.bind_group_layout, &self.instance_data_buffer, &texture_views)
         } else {
-            // We skip the first entity, because we don't want the player entity to show up
-            // in the picker buffer.
-            // TODO: Remove the player entity correctly.
-            for instruction in instructions.entities.iter().skip(player_size) {
+            for instruction in instructions.entities.iter().filter(|instruction| instruction.pickerable == true) {
```
hasenbanck, line 265:
```rust
            for instruction in instructions.entities.iter().filter(|instruction| instruction.pickerable) {
                self.draw_count += 1;
```

Like above.

korangar/src/graphics/passes/picker/entity.rs
```diff
@@ -229,10 +230,7 @@ impl Prepare for PickerEntityDrawer {
 
             let mut texture_views = Vec::with_capacity_in(self.draw_count, &self.bump);
```
hasenbanck, line 231:
```suggestion
            let mut texture_views = Vec::with_capacity_in(instructions.entities.len(), &self.bump);
```

korangar/src/graphics/instruction.rs
```diff
@@ -185,6 +185,7 @@ pub struct EntityInstruction {
     pub color: Color,
     pub mirror: bool,
     pub entity_id: EntityId,
+    pub pickerable: bool,
```
hasenbanck, line 188: Let's call this `add_to_picker`.

korangar/src/world/entity/mod.rs
```diff
@@ -1130,8 +1131,8 @@ impl Entity {
             .generate_pathing_mesh(device, queue, map, pathing_texture_mapping);
     }
 
-    pub fn render(&self, instructions: &mut Vec<EntityInstruction>, camera: &dyn Camera) {
-        self.get_common().render(instructions, camera);
+    pub fn render(&self, instructions: &mut Vec<EntityInstruction>, camera: &dyn Camera, pickerable: bool) {
```
hasenbanck, line 1134: Let's also call this `add_to_picker`

korangar/src/world/map/mod.rs
```diff
@@ -265,10 +265,11 @@ impl Map {
 
     #[cfg_attr(feature = "debug", korangar_debug::profile)]
     pub fn render_entities(&self, instructions: &mut Vec<EntityInstruction>, entities: &[Entity], camera: &dyn Camera, include_self: bool) {
-        entities
-            .iter()
-            .skip(!include_self as usize)
-            .for_each(|entity| entity.render(instructions, camera));
+        let mut is_current_player = true || include_self;
```
hasenbanck, line 268:
```rust
    #[cfg_attr(feature = "debug", korangar_debug::profile)]
    pub fn render_entities(&self, instructions: &mut Vec<EntityInstruction>, entities: &[Entity], camera: &dyn Camera) {
        entities.iter().enumerate().for_each(|(index, entity)| {
            entity.render(instructions, camera, index != 0);
        });
    }
```

We never call this function with "include_self==false"! So let's simplify it.

hasenbanck (changes requested):
Can you also please update the commit message?

`Explicitly encode if an entity should be included in the picker buffer`

korangar/src/graphics/passes/picker/entity.rs
```diff
@@ -199,7 +199,8 @@ impl Drawer<{ BindGroupCount::One }, { ColorAttachmentCount::One }, { DepthAttac
             let mut current_texture_id = self.solid_pixel_texture.get_id();
             pass.set_bind_group(2, self.solid_pixel_texture.get_bind_group(), &[]);
 
-            for (index, instruction) in draw_data[1..self.draw_count.saturating_add(1)].iter().enumerate() {
+            let instructions = draw_data.iter().filter(|instruction| instruction.add_to_picker);
+            for (index, instruction) in instructions.enumerate() {
```
kokosha, line 203: Find a bug in non-bind mode and changed to correct the behavior

hasenbanck (approved):

PR #168, opened 2024-12-01
On double right click, reset the camera rotation

vstumpf:
The worse way to implement a double-click handler.

### DIFF
--- a/korangar/src/input/mod.rs
+++ b/korangar/src/input/mod.rs
@@ -31,6 +31,12 @@ use crate::interface::resource::PartialMove;
 
 const MOUSE_SCOLL_MULTIPLIER: f32 = 30.0;
 const KEY_COUNT: usize = variant_count::<KeyCode>();
+const DOUBLE_CLICK_TIME_MS: u32 = 500;
+
+struct PreviousMouseButton {
+    button: MouseButton,
+    tick: ClientTick,
+}
 
 pub struct InputSystem {
     previous_mouse_position: ScreenPosition,
@@ -45,6 +51,7 @@ pub struct InputSystem {
     mouse_input_mode: MouseInputMode,
     input_buffer: Vec<char>,
     picker_value: Arc<AtomicU64>,
+    previous_mouse_button: PreviousMouseButton,
 }
 
 impl InputSystem {
@@ -63,6 +70,10 @@ impl InputSystem {
 
         let mouse_input_mode = MouseInputMode::None;
         let input_buffer = Vec::new();
+        let previous_mouse_button = PreviousMouseButton {
+            button: MouseButton::Left,
+            tick: ClientTick::new(0),
+        };
 
         Self {
             previous_mouse_position,
@@ -77,6 +88,7 @@ impl InputSystem {
             mouse_input_mode,
             input_buffer,
             picker_value,
+            previous_mouse_button,
         }
     }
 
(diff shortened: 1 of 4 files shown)


korangar/src/input/mod.rs
```diff
@@ -312,6 +316,18 @@ impl InputSystem {
             self.mouse_input_mode = MouseInputMode::RotateCamera;
         }
 
+        if self.right_mouse_button.pressed() && self.previous_mouse_button.0 == MouseButton::Right && client_tick.0.saturating_sub(self.previous_mouse_button.1.0) < DOUBLE_CLICK_TIME_MS {
+            events.push(UserEvent::CameraResetRotation);
+        }
+
+        if self.right_mouse_button.pressed() && !matches!(self.mouse_input_mode, MouseInputMode::RotateCamera) {
```
vstumpf, line 323: I'm trying to prevent a right-click-drag (when we rotate the camera) to count as the first click.

korangar/src/input/mod.rs
```diff
@@ -31,6 +31,7 @@ use crate::interface::resource::PartialMove;
 
 const MOUSE_SCOLL_MULTIPLIER: f32 = 30.0;
 const KEY_COUNT: usize = variant_count::<KeyCode>();
+const DOUBLE_CLICK_TIME_MS: u32 = 300;
```
vstumpf, line 34: Can be config-ed later?

korangar/src/world/cameras/player.rs
```diff
@@ -57,6 +58,10 @@ impl PlayerCamera {
         self.view_angle.move_desired(rotation * ROTATION_SPEED);
     }
 
+    pub fn reset_rotate(&mut self) {
```
hasenbanck, line 61:
```suggestion
    pub fn reset_rotation (&mut self) {
```

I think reset_rotation would be a better name.

korangar/src/input/mod.rs
hasenbanck: The default doube click time-frame on windows is 500 ms. Since most poeple most likely have this in their muscle memory, we should leave it as 500 ms too.

[Source](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setdoubleclicktime)

korangar/src/input/mod.rs
```diff
@@ -45,6 +46,7 @@ pub struct InputSystem {
     mouse_input_mode: MouseInputMode,
     input_buffer: Vec<char>,
     picker_value: Arc<AtomicU64>,
+    previous_mouse_button: (MouseButton, ClientTick),
```
hasenbanck, line 49: Can you please include an intermediate struct that has descriptive names for the fields?

hasenbanck (changes requested):
Thank you for the PR!

korangar/src/input/mod.rs
vstumpf: google ai failed me

korangar/src/input/mod.rs
hasenbanck: Testing it locally, I think it works fine. Do you think the current approach need further refinment?

hasenbanck:
Can you please squash the commits into one and also rebase the branch against current main?

hasenbanck (approved):

PR #179, opened 2024-12-06
Mute audio on unfocus

kokosha:
Add an audio setting to mute the audio on windows unfocus.

### DIFF
--- a/korangar/src/audio/settings.rs
+++ b/korangar/src/audio/settings.rs
@@ -1,0 +1,48 @@
+#[cfg(feature = "debug")]
+use korangar_debug::logging::{print_debug, Colorize};
+use ron::ser::PrettyConfig;
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Serialize, Deserialize)]
+pub struct AudioSettings {
+    pub mute_on_focus_loss: bool,
+}
+
+impl Default for AudioSettings {
+    fn default() -> Self {
+        Self { mute_on_focus_loss: true }
+    }
+}
+
+impl AudioSettings {
+    const FILE_NAME: &'static str = "client/audio_settings.ron";
+
+    pub fn new() -> Self {
+        Self::load().unwrap_or_else(|| {
+            #[cfg(feature = "debug")]
+            print_debug!("failed to load audio settings from {}", Self::FILE_NAME.magenta());
+            Default::default()
+        })
+    }
+
+    pub fn load() -> Option<Self> {
+        #[cfg(feature = "debug")]
+        print_debug!("loading audio settings from {}", Self::FILE_NAME.magenta());
+        std::fs::read_to_string(Self::FILE_NAME)
+            .ok()
+            .and_then(|data| ron::from_str(&data).ok())
+    }
+
+    pub fn save(&self) {
+        #[cfg(feature = "debug")]
+        print_debug!("saving audio settings to {}", Self::FILE_NAME.magenta());
+        let data = ron::ser::to_string_pretty(self, PrettyConfig::new()).unwrap();
+        std::fs::write(Self::FILE_NAME, data).expect("unable to write file");
+    }
+}
+
+impl Drop for AudioSettings {
+    fn drop(&mut self) {
+        self.save();
+    }
+}
--- a/korangar/src/main.rs
+++ b/korangar/src/main.rs
@@ -24,6 +24,7 @@ macro_rules! time_phase {
     }
 }
 
+mod audio;
 mod graphics;
 mod input;
 #[macro_use]
@@ -80,6 +81,7 @@ use winit::event_loop::{ActiveEventLoop, ControlFlow, EventLoop};
 use winit::keyboard::PhysicalKey;
 use winit::window::{Icon, Window, WindowId};
 
+use crate::audio::*;
 use crate::graphics::*;
 use crate::input::{InputSystem, UserEvent};
 use crate::interface::application::InterfaceSettings;
@@ -196,6 +198,8 @@ struct Client {
     high_quality_interface: MappedRemote<GraphicsSettings, bool>,
     #[cfg(feature = "debug")]
     render_settings: PlainTrackedState<RenderSettings>,
+
+    mute_on_focus_loss: MappedRemote<AudioSettings, bool>,
 
     application: InterfaceSettings,
     interface: Interface<InterfaceSettings>,
@@ -351,6 +355,9 @@ impl Client {
         });
 
         time_phase!("create audio engine", {
+            let audio_settings = PlainTrackedState::new(AudioSettings::new());
+            let mute_on_focus_loss = audio_settings.mapped(|settings| &settings.mute_on_focus_loss).new_remote();
+
             let audio_engine = Arc::new(AudioEngine::new(game_file_loader.clone()));
         });
 
@@ -593,6 +600,7 @@ impl Client {
             high_quality_interface,
             #[cfg(feature = "debug")]
             render_settings,
+            mute_on_focus_loss,
             application,
             interface,
             focus_state,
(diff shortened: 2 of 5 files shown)


korangar_audio/src/settings.rs
```diff
@@ -0,0 +1,52 @@
+#[cfg(feature = "debug")]
```
hasenbanck, line 1: Please don't move the settings into the audio crate. It should be agnostic of application specific code.

The graphic engine currently only has the "GraphicsSettings", because we haven't move it to it's own crate yet.

korangar_audio/src/settings.rs
```diff
@@ -0,0 +1,52 @@
+#[cfg(feature = "debug")]
+use korangar_debug::logging::{print_debug, Colorize};
+use ron::ser::PrettyConfig;
+use serde::{Deserialize, Serialize};
+/// Audio Settings
```
hasenbanck, line 5: Please also properly format the file.

korangar_audio/src/lib.rs
```diff
@@ -191,8 +194,57 @@ impl<F: FileLoader> AudioEngine<F> {
...
+    }
+
+    /// This function activate during windows resume state
+    pub fn on_resume(&self) {
+        self.mute(false);
+    }
 
-        AudioEngine { engine_context }
+    /// This function activate during windows suspend state
+    pub fn on_suspended(&self) {
```
hasenbanck, line 219: The audio code doesn't need to know about the focus state of a windows. A "mute" function should be enough.

korangar/src/main.rs
```diff
@@ -2417,5 +2437,6 @@ impl ApplicationHandler for Client {
 
     fn suspended(&mut self, _event_loop: &ActiveEventLoop) {
         self.graphics_engine.on_suspended();
+        self.audio_engine.on_suspended();
```
hasenbanck, line 2440: The graphic engine only has a "on_suspend()" because it needs to be tightly integrated into the windows / graphic output loop. No need to couple the audio engine too.

hasenbanck (changes requested):

korangar_audio/src/settings.rs
kokosha: I am going to move the files to korangar/audio/settings.

korangar_audio/src/lib.rs
kokosha: I will create a audio window state to handle the window state interaction.

korangar/src/main.rs
kokosha: The on_suspend depend on the operating system, in some operating system may continue to reproduce the sound during on_suspend. Maybe on Wayland, X11, I am not sure. If you want, I can remove it.

korangar_audio/src/lib.rs
```diff
@@ -191,8 +192,29 @@ impl<F: FileLoader> AudioEngine<F> {
...
             sound_effect_track,
         });
+        AudioEngine {
+            engine_context,
+            mute_on_focus_loss: Mutex::new(true),
+        }
+    }
+
+    /// This function set the bool mute_on_focus_loss
+    pub fn set_mute_on_focus_loss(&self, mute_on_focus_loss: bool) {
```
hasenbanck, line 202: We don't need the functions that use the mutex anymore.

korangar_audio/src/lib.rs
```diff
@@ -191,8 +192,29 @@ impl<F: FileLoader> AudioEngine<F> {
...
+    }
 
-        AudioEngine { engine_context }
+    /// This function is used to set the settings during audio engine creation
+    pub fn on_create(&self, mute_on_focus_loss: bool) {
+        self.set_mute_on_focus_loss(mute_on_focus_loss);
+    }
+
+    /// This function mute or unmute the audio
+    pub fn mute(&self, mute_audio: bool) {
```
hasenbanck, line 212:
```suggestion
    pub fn mute(&self, enable: bool) {
```

korangar_audio/src/lib.rs
```diff
@@ -191,8 +192,29 @@ impl<F: FileLoader> AudioEngine<F> {
...
+        *self.mute_on_focus_loss.lock().unwrap() = mute_on_focus_loss;
+    }
 
-        AudioEngine { engine_context }
+    /// This function is used to set the settings during audio engine creation
+    pub fn on_create(&self, mute_on_focus_loss: bool) {
+        self.set_mute_on_focus_loss(mute_on_focus_loss);
+    }
+
+    /// This function mute or unmute the audio
```
hasenbanck, line 211:
```suggestion
    /// Mutes or unmutes the audio.
```

korangar/src/audio/settings.rs
```diff
@@ -0,0 +1,53 @@
+#[cfg(feature = "debug")]
+use korangar_debug::logging::{print_debug, Colorize};
+use ron::ser::PrettyConfig;
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Serialize, Deserialize)]
+pub struct AudioSettings {
+    // Mute audio on focus loss
```
hasenbanck, line 8: This comment is not very helpfull. I think you can remove it.

korangar/src/audio/audio_state.rs
```diff
@@ -0,0 +1,40 @@
+use std::sync::Arc;
+
+use korangar_audio::AudioEngine;
+
+use crate::GameFileLoader;
+pub struct AudioWindowState {
```
hasenbanck, line 6: I don't think you need this state at all. You can just call the audio engine functions directly.

hasenbanck (changes requested):

korangar/src/main.rs
hasenbanck: That's not what I tried to say. I just said that the audio engine itself doesn't need to know the concept of "a windows suspended". That is something the application can handle and call the audio engine functionality directly.

korangar/src/audio/audio_state.rs
kokosha: I think it is better to create a struct that handles the audio settings states but not sure how to call the function and to not directly put this audio function inside the main loop.

hasenbanck: But look how much it currently does (next to nothing). So this abstraction is not needed yet!

korangar/src/main.rs
```diff
@@ -2357,8 +2366,10 @@ impl ApplicationHandler for Client {
                 *self.ssaa.get(),
                 *self.screen_space_anti_aliasing.get(),
                 *self.high_quality_interface.get(),
-            )
+            );
         }
+
+        self.audio_engine.mute(false);
```
hasenbanck, line 2372: You should only unmute it if "mute_on_focus_loss" is set!

korangar/src/main.rs
```diff
@@ -2384,6 +2395,12 @@ impl ApplicationHandler for Client {
                 if !focused {
                     self.input_system.reset();
                     self.focus_state.remove_focus();
+                    let mute_on_focus_loss = *self.mute_on_focus_loss.get();
+                    if mute_on_focus_loss {
+                        self.audio_engine.mute(true);
+                    }
+                } else {
+                    self.audio_engine.mute(false);
```
hasenbanck, line 2403: You should only unmute it if "mute_on_focus_loss" is set!

So better move the mute function out of the `if !focused` and instead write:

```rust
let mute_on_focus_loss = *self.mute_on_focus_loss.get();
if mute_on_focus_loss {
    self.audio_engine.mute(!focused);
}
```

```

File: korangar/src/main.rs (hunk 4 above)
Comment (kokosha) [reply korangar/src/main.rs]: Oh, I misunderstand

File: korangar/src/audio/settings.rs [hunk 12]
```diff
@@ -0,0 +1,52 @@
...
+    /// Save audio settings
+    pub fn save(&self) {
+        #[cfg(feature = "debug")]
+        print_debug!("saving audio settings to {}", Self::FILE_NAME.magenta());
+        let data = ron::ser::to_string_pretty(self, PrettyConfig::new()).unwrap();
+        std::fs::write(Self::FILE_NAME, data).expect("unable to write file");
+    }
+}
+
+/// Drop audio settings
```
hasenbanck, line 47: Oh, I didn't saw this before: You don't need to comment the impl block of Drop.

hasenbanck (changes requested):
I think this is the last review cycle we need!

korangar/src/main.rs
kokosha: If you mute with mute_on_focus_loss activated and later you deactivate mute_on_focus_loss during the mute, the audio will stay muted.

hasenbanck: But here we handle the case where the window if focuses again. You can't change the "mute_on_focus_loss" option while you are unfocused!

kokosha: True. Theoretically it is possible.

hasenbanck:
Please rebase. After that, I will merge the PR.

hasenbanck (approved):

PR #184, opened 2024-12-07
Change entity bounding-box per action instead of every action

kokosha:
This is a small change in logic that fix the rendering of monster that have some action with underground sprites, and not let the error to affect the other action.

### DIFF
--- a/korangar/src/loaders/animation/mod.rs
+++ b/korangar/src/loaders/animation/mod.rs
@@ -201,63 +201,63 @@ impl AnimationLoader {
             animations.push(Animation { frames });
         }
 
-        // The problem is that each frame of an action is not the same size
-        // and without size consistency, the proportion differ between frames,
-        // causing pixel offsets.
-        // To resolve this, we resized the frames to ensure
-        // identical frame size and origin point at (0,0) between frames.
-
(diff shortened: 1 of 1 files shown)


hasenbanck:
Should we leave this open as a draft until you found the correct math?

kokosha:
This is a small change in logic that fix the rendering of monster that have some action with underground sprites, and not let the error to affect the other action.
Maybe the comment is not well-explained.
The underground sprite is going to be fixed in another pull request, when I get the math correct.

korangar/src/loaders/animation/mod.rs
```diff
@@ -201,63 +201,62 @@ impl AnimationLoader {
...
+            // To resolve this, we resized the frames to ensure
+            // identical frame size and origin point at (0,0) between frames.
+
+            // First, identify the bounding box that encompasses all motion
+            // frames.
+            let mut min_top = i32::MAX;
+            let mut max_bottom = 0;
+            let mut min_left = i32::MAX;
+            let mut max_right = 0;
+            // The player can change the sprite and causes an offset from a image to
```
hasenbanck, line 218: Can you please add a empty line before the comment?

kokosha:
I think it is ok

PR #197, opened 2024-12-16
Remove unused Writable::create() function

hasenbanck:
The functionality isn't used anymore, since the add_file() function will call create_dir_all() if not all folders are found.

### DIFF
--- a/korangar/src/loaders/archive/folder/mod.rs
+++ b/korangar/src/loaders/archive/folder/mod.rs
@@ -70,11 +70,6 @@ impl Archive for FolderArchive {
 }
 
 impl Writable for FolderArchive {
-    fn create(&mut self) {
-        fs::create_dir_all(&self.folder_path)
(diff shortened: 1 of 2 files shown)


vE5li:
Can we make this a `cfg_attr` only when `patched_as_folder` is enabled?

hasenbanck:
Sure! Great idea!

hasenbanck:
@vE5li After having looked intot he code: Where do we use this anyhow? I can't find any usage of the create function in the code, so putting it behind "patched_as_folder" seems odd right now.

vE5li:
Huh.. I was sure it's used in the gamefile loader when creating the patched lua files but I can't find it either (I'm on mobile so take that with a grain of salt). Maybe it's not used anymore.

hasenbanck:
We use "create_dir_all" in the "add_file" also, so it seems we don't need a dedicated logic to create the whole folder structure anymore. I will remove the function for now.

PR #203, opened 2024-12-19
Implement MSDF font rendering

hasenbanck:


kokosha:
Screenshot 2024-12-20 184505
The chat message offset seems not correct in the above image. When you @warp and later say anything, the offset seems wrong. It may be caused by the missing character space ` ` in the warp message and this change the offset of the message, because it doesn't contain the character space ` `, making the offset wrong.

korangar/src/renderer/interface.rs
```diff
@@ -146,10 +146,10 @@ impl korangar_interface::application::InterfaceRenderer<InterfaceSettings> for I
             font_size = font_size * 2.0;
         }
 
-        let TextLayout { glyphs, mut size } =
+        let TextLayout { glyphs, mut size, .. } =
             self.font_loader
```
kokosha, line 150: I think it is calculated here, not sure

eleriaqueen:
There's also an issue with wrapped text which is not present in main branch.
Screenshot_20241221_082736
Screenshot_20241221_082754

hasenbanck:
It seems the text_height is not calculated correctly. This is slightly confusing:

```rust
let x = physical_glyph.x as f32 + glyph_coordinate.offset_left;
let y = run.line_y + physical_glyph.y as f32 - glyph_coordinate.offset_top;
let width = glyph_coordinate.width;
let height = glyph_coordinate.height;

text_height = text_height.max(y + height);

let position = Rectangle::new(Point2::new(x, y), Point2::new(x + width, y + height));
let color = layout_glyph.color_opt.map(|color| color.into()).unwrap_or(default_color);

glyphs.push(GlyphInstruction {
    position,
    texture_coordinate: glyph_coordinate.texture_coordinate,
    color,
});
```

We use the x/y/width/height also for the glyph position. Since the glyphs are layed out correct, I wonder why the text height is incorrect. Maybe the glyph stores now offets in a different "direction"? I have to debug this more to understand the true cause for this (we didn't change the layout rendering itself in a fundamental way, only the "gylph storage").

hasenbanck:
@eleriaqueen @kokosha I think I fixed the issue. I took out the glyphs out of the equation and use the line height directly. This should fix the wrapping issues you have encountered. Would be happy if you could test again.

kokosha:
It is working now!

eleriaqueen:
Yep looks like you fixed the issue.

PR #204, opened 2024-12-20
Set the encoding to UTF-8 when reading string with FromBytes

kokosha:
Change the encoding to UTF-8 when reading string, and also fall back to single-byte character encoding for unsupported string encodings.
The issue discovered is that the string is being processed one byte at a time. However, in UTF-8 encoding, characters can be represented by two or more bytes, which might cause encoding issues.
For example the character á is encoded incorrectly.

### DIFF
--- a/ragnarok_bytes/src/from_bytes/implement.rs
+++ b/ragnarok_bytes/src/from_bytes/implement.rs
@@ -65,16 +65,19 @@ impl<T: FromBytes, const SIZE: usize> FromBytes for [T; SIZE] {
 
 impl FromBytes for String {
     fn from_bytes<Meta>(byte_reader: &mut ByteReader<Meta>) -> ConversionResult<Self> {
-        let mut value = String::new();
+        let mut value = Vec::<u8>::new();
 
         while let Ok(byte) = byte_reader.byte::<Self>() {
             match byte {
                 0 => break,
-                byte => value.push(byte as char),
+                byte => value.push(byte),
             }
         }
+        let result = String::from_utf8(value)
+            .map_err(|error| error.into_bytes())
+            .unwrap_or_else(|error| error.iter().map(|byte| *byte as char).collect());
 
-        Ok(value)
+        Ok(result)
     }
 }
 


vE5li:
The encoding should be fixed before creating the Network event already. Ideally, for all packets that use UTF-8 encoded strings.

hasenbanck:
Is this PR still in draft status?

kokosha:
I think it can be merged.

ragnarok_bytes/src/from_bytes/implement.rs
```diff
@@ -65,16 +65,20 @@ impl<T: FromBytes, const SIZE: usize> FromBytes for [T; SIZE] {
...
+        let mut value = Vec::<u8>::new();
 
         while let Ok(byte) = byte_reader.byte::<Self>() {
             match byte {
                 0 => break,
-                byte => value.push(byte as char),
+                byte => value.push(byte),
             }
         }
+        let result = match String::from_utf8(value).map_err(|error| error.into_bytes()) {
```
hasenbanck, line 76:
```suggestion
        let result = String::from_utf8(value)
            .map_err(|error| error.into_bytes())
            .unwrap_or_else(|error| error.iter().map(|byte| *byte as char).collect());
```

My linter proposes the following version.

hasenbanck (changes requested):

ragnarok_bytes/src/from_bytes/implement.rs
kokosha: Wow, it seems syntactically better.

hasenbanck: Just as a heads up: You actually didn't pushed the change, I think.

hasenbanck:
@vE5li Do you think that the issue is now fixed in the correct place now?

vE5li:
This would be ideal! Question is if the loaders can still properly find everything. If everything still works as expected I would merge it, worst case we introduce a second string type later on.

hasenbanck:
> This would be ideal! Question is if the loaders can still properly find everything. If everything still works as expected I would merge it, worst case we introduce a second string type later on.

It seems to work as expected. I will merge for now and make a mental note that things might could break because of this change.

hasenbanck (approved):

PR #205, opened 2024-12-20
Add missing feature debug tag on hidden element in linked list

kokosha:


korangar/src/interface/linked.rs
```diff
@@ -94,9 +97,11 @@ mod test {
     fn links_can_be_hidden() {
         let element = LinkedElement::new();
 
+        #[cfg(feature = "debug")]
```
hasenbanck, line 100: The test tests `link_hidden()`! So the whole test should get the cfg!

hasenbanck (changes requested):

korangar/src/interface/linked.rs
kokosha: Good catch!

hasenbanck (approved):

PR #206, opened 2024-12-21
Implement custom path finding

hasenbanck:
I found the path finding code highly complicated and also inefficient, since we allocated on the heap multiple times, without amortising the allocations.

This implementation provides a struct that implements the A* and also line of sight path finding in a way, that is compatible with the server implementations. We also re-use the heap allocations.

The A* algorithm is a basic implementation that uses a binary heap. Moved the code in korangar_utils, so that it can be properly optimized and also tested.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -688,18 +688,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "deprecate-until"
-version = "0.1.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7a3767f826efbbe5a5ae093920b58b43b01734202be697e1354914e862e8e704"
-dependencies = [
- "proc-macro2",
- "quote",
- "semver",
- "syn",
-]
-
-[[package]]
 name = "derive-new"
 version = "0.7.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


eleriaqueen:
I just tested your PR a minute ago.
Is it possible that moving is bit too slow when going in a straight line, and maybe a bit too fast when doing diagonals ?
Because it feels that way.

hasenbanck:
> I just tested your PR a minute ago. Is it possible that moving is bit too slow when going in a straight line, and maybe a bit too fast when doing diagonals ? Because it feels that way compared to official.

Yes, that is the issue that I talked about in the last paragraph. But fixed that issue in my local branch allready and will upload the fix soon.

hasenbanck:
Should be fixed now and the behavior should be no worse than the old implementation.

eleriaqueen:
Yep fixed, and movement feels pretty good ! 

There is just a little issue that was present in main as well and might be unrelated but when first appearing into the world after character selection, client is probably still loading assets then because movement is super jittery for about 5 seconds.

hasenbanck:
> Yep fixed, and movement feels pretty good !
> 
> There is just a little issue that was present in main as well and may be unrelated but when first appearing into the world after character selection, client is probably still loading assets then because movement is super jittery for about 5 seconds.

I'm not too sure about that. The only thing that could load after you spawned are entities, but even if I spawn in an empty location where no entities are, I get those "desyncs". I think the server doesn't expect the client to move so fast, maybe we need to wait for a specific packet to arrive, until we can send out player commands. That issue is for future investigations though.

PR #211, opened 2024-12-27
Change the direction of the entities

kokosha:


ragnarok_packets/src/position.rs
```diff
@@ -52,10 +57,18 @@ impl WorldPosition2 {
...
     pub fn to_origin_destination(self) -> (WorldPosition, WorldPosition) {
-        (WorldPosition { x: self.x1, y: self.y1 }, WorldPosition {
-            x: self.x2,
-            y: self.y2,
-        })
+        (
+            WorldPosition {
+                x: self.x1,
+                y: self.y1,
+                direction: 0,
```
hasenbanck, line 64: Does this packet not have a direction in it? Please have a look at the `from_bytes()` function for this packet. It seems it was planned to read a direction there too. If that is not the case, then please remove that comment in there.

ragnarok_packets/src/position.rs
```diff
@@ -5,11 +5,12 @@ use ragnarok_bytes::{ByteReader, ConversionResult, FromBytes, ToBytes};
 pub struct WorldPosition {
     pub x: usize,
     pub y: usize,
+    pub direction: usize,
```
hasenbanck, line 8: Couldn't we create a type for this? I'm not sure which number has which direction, so you would need to double-check that.

```rust
#[derive(Debug, Copy, Clone, ByteConvertable)]
#[cfg_attr(feature = "interface", derive(korangar_interface::elements::PrototypeElement))]
pub enum Direction {
    N = 0,
    NE = 1,
    E = 2,
    SE = 3,
    S = 4,
    SW = 5,
    W = 6,
    NW = 7,
}

impl From<Direction> for usize {
    fn from(value: Direction) -> Self {
        value as usize
    }
}

impl From<usize> for Direction {
    fn from(value: usize) -> Self {
        match value {
            1 => Direction::NE,
            2 => Direction::E,
            3 => Direction::SE,
            4 => Direction::S,
            5 => Direction::SW,
            6 => Direction::W,
            7 => Direction::NW,
            0 | _ => Direction::N,
        }
    }
}
```

hasenbanck (changes requested):

ragnarok_packets/src/position.rs
hasenbanck: The `From<usize>` will be optimized to:

```asm
mov     rax, rdi
and     al, 7
ret
```

korangar/src/world/animation/mod.rs
```diff
@@ -82,10 +82,11 @@ impl Default for AnimationFramePart {
 }
 
 impl AnimationData {
-    pub fn get_frame(&self, animation_state: &AnimationState, camera: &dyn Camera, head_direction: usize) -> &AnimationFrame {
+    pub fn get_frame(&self, animation_state: &AnimationState, camera: &dyn Camera, direction: Direction) -> &AnimationFrame {
         let camera_direction = camera.camera_direction();
-        let direction = (camera_direction + head_direction) % 8;
-        let animation_action_index = animation_state.action as usize * 8 + direction;
+        let mut direction_usize: usize = direction.into();
+        direction_usize = (camera_direction + direction_usize) & 7;
```
hasenbanck, line 88:
```suggestion
        let direction_usize = (camera_direction + usize::from(direction)) & 7;
```

korangar/src/world/entity/mod.rs
```diff
@@ -449,7 +452,7 @@ impl Common {
 
                 let array = last_step_position - next_step_position;
                 let array: &[isize; 2] = array.as_ref();
-                self.head_direction = match array {
+                self.direction = match array {
                     [0, 1] => 0,
```
hasenbanck, line 456: You can use the `Direction` enum here now! To remove these magic values is the reason why we added it!

ragnarok_packets/src/position.rs
```diff
@@ -19,19 +65,26 @@ impl FromBytes for WorldPosition {
...
+            direction: direction.into(),
+        })
     }
 }
 
 impl ToBytes for WorldPosition {
     fn to_bytes(&self) -> ConversionResult<Vec<u8>> {
         let mut coordinates = vec![0, 0, 0];
+        let mut direction: usize = self.direction.into();
+        direction = (8 - direction + 4) & 7;
```
hasenbanck, line 83:
```suggestion
        let direction = (8 - usize::from(self.direction) + 4) & 7;
```

ragnarok_packets/src/position.rs
kokosha: Not sure if I used into() in the way it should be used.

ragnarok_packets/src/position.rs
kokosha: I have tried to understand the reasoning, but seems not related with direction.

ragnarok_packets/src/position.rs
```diff
@@ -67,7 +128,9 @@ impl FromBytes for WorldPosition2 {
         let y1 = (coordinates[2] >> 4) | ((coordinates[1] & 0b111111) << 4);
         let x2 = (coordinates[3] >> 2) | ((coordinates[2] & 0b1111) << 6);
         let y2 = coordinates[4] | ((coordinates[3] & 0b11) << 8);
-        //let direction = ...
+        /* Not related with direction at all is bigger than 7, it seems to be always 8.
```
hasenbanck, line 131: If you do this, then you can add one `unknown` u8 field for the element, right? But have you checked other implementations for it's possible meaning?

hasenbanck (changes requested):

ragnarok_packets/src/position.rs
kokosha: I checked it seems a shift from x1, y1, but it seems that until now the shift is always 0.

hasenbanck: It seems to be a height "level" information? We should at least read the field and call it "unknown" for now.

ragnarok_packets/src/position.rs
```diff
@@ -1,15 +1,83 @@
...
+            4 => Direction::S,
+            5 => Direction::SW,
+            6 => Direction::W,
+            7 => Direction::NW,
+            _ => unreachable!(),
+        }
+    }
+}
+
+impl From<&[isize; 2]> for Direction {
```
hasenbanck, line 40: At least only implement the type on the non reference type! De-reference the type before calling the into function.

ragnarok_packets/src/position.rs
hasenbanck: Please also remove the comment at the "world_position_2" test case, since world_position_2 seems to have no orientation at all.

kokosha: Ok, I will do that.

kokosha: Don't know if it what I am thinking.

hasenbanck:
Please squash so that I can merge it.

hasenbanck (approved):

PR #214, opened 2024-12-28
Improve first map server connect

hasenbanck:
On first login the map server will send a command to change the map regardless, so we don't need to handle the map loading in the network event when we disconnect from the character server. We only initialize the player entity and interface in this case. We also now properly turn the screen black until we fully finished loading the map. Later we should also do the loading asynchronously on map change and turn the screen black while loading.

### DIFF
--- a/korangar/src/main.rs
+++ b/korangar/src/main.rs
@@ -274,7 +274,7 @@ struct Client {
     chat_messages: PlainTrackedState<Vec<ChatMessage>>,
     main_menu_click_sound_effect: SoundEffectKey,
 
-    map: Map,
+    map: Option<Map>,
 }
 
 impl Client {
@@ -679,7 +679,7 @@ impl Client {
             tile_texture,
             chat_messages,
             main_menu_click_sound_effect,
-            map,
+            map: Some(map),
         }
     }
 
@@ -853,7 +853,7 @@ impl Client {
                     self.point_light_manager.clear();
                     self.audio_engine.play_background_music_track(None);
 
-                    self.map = self
+                    let map = self
                         .map_loader
                         .load(
                             DEFAULT_MAP.to_string(),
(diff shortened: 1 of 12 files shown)


korangar/src/main.rs
```diff
@@ -938,21 +942,6 @@ impl Client {
                         .cloned()
                         .unwrap();
 
-                    self.map = self
```
kokosha, line 941: The self.map information is used in set_position from NetworkEvent::SetPlayerPosition.
If you are outside of the boundary of the start screen map, the games crashes.

hasenbanck: I see, we sometimes convert one network packet into multiple network events, which is ... suboptimal. Yes, the server ist sending also a position when the login "MapServerLoginSuccessPacket" was received, but every ChangeMapPacketPacket will generate an Event, that contains the map name and the position. I think it's best to alwas only send out one NetworkEvent for one packet.

The old way we did it:

CharacterSelectionSuccessPacket -> NetworkEvent::CharacterSelected -> load player, change map, set player at origin
Connect to Map server -> MapServerLoginSuccessPacket -> NetworkEvent::UpdateClientTick + NetworkEvent::SetPlayerPosition -> Set player at correct position  + update client tick
ChangeMapPacket -> NetworkEvent::ChangeMap -> changes map again, set player at correct position


I propose that we simplify the first sparn proces:


CharacterSelectionSuccessPacket -> NetworkEvent::CharacterSelected -> load player, set player at origin
MapServerLoginSuccessPacket -> NetworkEvent::UpdateClientTick -> update client tick
ChangeMapPacket -> NetworkEvent::ChangeMap -> changes map forst the first time, set player at correct position


We essentially now only allow that one packet can create exactly one network event. We also don't use the map information in the CharacterSelectionSuccessPacket , ignore the position in the MapServerLoginSuccessPacket.
But instead only use the map that the map server tolds us to change into and also the position it gaves us.

PR #248, opened 2025-02-09
Fix pop sound on Wayland when the client opens

vE5li:


hasenbanck:
This is a working fix for https://github.com/tesselode/kira/issues/121
We defer the loading of spatial sound sources (called ambient sound source in our source code), to when the window is visible / focused. If we load them when the window is not visible, we get the "pop" sound under linux (but not windows or mac).

PR #249, opened 2025-02-10
Implement PCSS shadows

hasenbanck:


korangar/src/settings/graphic.rs
```diff
@@ -29,13 +29,13 @@ impl Default for GraphicsSettings {
             lighting_mode: LightingMode::Enhanced,
             vsync: true,
             limit_framerate: LimitFramerate::Unlimited,
-            triple_buffering: true,
+            triple_buffering: false,
```
vE5li, line 32: Is this change intentional?

hasenbanck: No. Must be a leftover of me testing a lot.

PR #251, opened 2025-02-12
Fix broken ambient lights

hasenbanck:
Some maps, like "yuno", have ambient lights that outshine all shadows. This workaround detect such cases and sets a saner value.


vE5li:
I am not a fan of this solution. It breaks as soon as we get a map with `1.0 0.95 1.0 1.0`.

PR #257, opened 2025-02-20
Update cargo toolchain and update edition to 2024

hasenbanck:
Changes were made with:

 - cargo fix --edition
 - cargo fmt --all
 
 Also fixed some compiler errors and obvious lints.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,6 @@ korangar/[Bb][Gg][Mm]/
 
 # macos
 .DS_Store
+
+# Idea
+.idea/
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,5 @@
 [workspace]
-resolver = "2"
+resolver = "3"
 members = ["korangar", "ragnarok_*", "korangar_*"]
 
 [workspace.dependencies]
--- a/korangar/Cargo.toml
+++ b/korangar/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "korangar"
 version = "0.1.0"
-edition = "2021"
+edition = "2024"
 
 [dependencies]
 arrayvec = { workspace = true }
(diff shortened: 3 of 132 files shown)


vE5li:
Two questions:
- Are the changes to the `.gitignore` from you or was that also automatic?
- There is a one-liner if now (`if mip_level_count > 1 { 2 << (mip_level_count - 1) } else { 2 }`) now, is that because of the new edition?

hasenbanck:
>     * Are the changes to the `.gitignore` from you or was that also automatic?

Mine, since you need to have a "clean" git repository for running `cargo fix --eddition`. We also have stuff for Nix, so I guess adding stuff for Idea is not a big issue.
 
>     * There is a one-liner if now (`if mip_level_count > 1 { 2 << (mip_level_count - 1) } else { 2 }`), is that because of the new edition?

It's most likely because of the new style edition yes. `cargo fmt --all` does this. I normaly don't deviate from the standard tool formatting.

vE5li:
> Mine, since you need to have a "clean" git repository for running cargo fix --eddition. We also have stuff for Nix, so I guess adding stuff for Idea is not a big issue.

It's not an issue, I was just curious.

> [...] I normaly don't deviate from the standard tool formatting.

I mean we are already deviating from the standard formatting. I wil try to keep an open mind about this one liner and see how it feels over time. If it bothers me too much (mainly because I think it reduces readability) I might change it in a separate PR.

kokosha:
I reviewed the PR, everything seems fine, only these two things are bothering me:

1. if let changing to match directly in:
korangar/src/main.rs
korangar/src/input/mod.rs
korangar_audio/src/lib.rs
korangar_util/src/container/generational_slab.rs
korangar_util/src/container/simple_slab.rs
ragnarok_procedural/src/helper.rs

2. inline if else
if mip_level_count > 1 { 2 << (mip_level_count - 1) } else { 2 }
if sun_intensity > moon_intensity { (1.0, 0.0) } else { (0.0, 1.0) }

PR #261, opened 2025-02-21
Fix for strange artifacts when not using high-quality interface

kokosha:
The bug affects users in Vulkan using AMD GPU on a Windows platform.

### DIFF
--- a/korangar/src/graphics/engine.rs
+++ b/korangar/src/graphics/engine.rs
@@ -858,352 +858,355 @@ impl GraphicsEngine {
         let mut forward_encoder = self.device.create_command_encoder(&CommandEncoderDescriptor::default());
         let mut post_processing_encoder = self.device.create_command_encoder(&CommandEncoderDescriptor::default());
 
-        rayon::in_place_scope(|scope| {
-            // Picker Pass
-            scope.spawn(|_| {
-                let mut render_pass =
-                    engine_context
-                        .picker_render_pass_context
-                        .create_pass(&mut picker_encoder, &engine_context.global_context, None);
-
-                if let Some(map_picker_tile_vertex_buffer) = instruction.map_picker_tile_vertex_buffer.as_ref() {
-                    engine_context
-                        .picker_tile_drawer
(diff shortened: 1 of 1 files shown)


hasenbanck:
You seem to have effectively shuffled the execution order. I will take a deeper look into the issue. I think we need to reduce the amount of parralization on windows with AMD drivers instead.

kokosha:
My intuition to that change is that rayon::in_place_scope run in series, while scope.spawn run in parallel.
In the old code, what happened is that scope.spawn activated the asynchronous code in series, but on the end the code still continuously execute the code inside rayon::in_place_scope, but for code that is not in scope.spawn.

kokosha:
I reproduced the bug. The first code is wrong, while the second code is right.
```rust
rayon::in_place_scope(|scope_master| {
    scope_master.spawn(|_| {
        for i in 0..5 {
            println!("worker 1");
        }
    });
    scope_master.spawn(|_| {
        for i in 0..5 {
            println!("worker 2");
        }
    });
    for i in 0..5 {
        println!("worker 3");
    }
});
```
```rust
rayon::in_place_scope(|scope_master| {
    rayon::in_place_scope(|scope| {
        scope.spawn(|_| {
            for i in 0..5 {
                println!("worker 1");
            }
        });
        scope.spawn(|_| {
            for i in 0..5 {
                println!("worker 2");
            }
        });
    });
    scope_master.spawn(|_| {
        for i in 0..5 {
            println!("worker 3");
        }
    });
});
```
The first code output
```txt
worker 3
worker 3
worker 3
worker 3
worker 2
worker 2
worker 1
worker 1
worker 3
worker 2
worker 2
worker 1
worker 1
worker 2
worker 1
```
The second code output
```txt
worker 1
worker 1
worker 1
worker 1
worker 2
worker 2
worker 2
worker 2
worker 2
worker 1
worker 3
worker 3
worker 3
worker 3
worker 3
```

hasenbanck:
Fixed instead by #263

Issue #165, opened 2024-11-29
Model animations played not found in the original client


hasenbanck:
Some models currently are animated, even though they are not in the original client.

It seems the specific magician building here in geffen shows an animation, that is also either broken in our client, or broken in the original file.

Original client:

https://github.com/user-attachments/assets/987b1a1f-102c-4594-94ab-65aeee87f4f7

Current main branch:

https://github.com/user-attachments/assets/87e51291-a2d7-46ec-a4f8-406dc200ac89

Cyanide0210:
Haha that looks funny, I think it should be stuck to the thing thats moving.. not sure if it should rotate tho

PR #264, opened 2025-02-23
Fix RSM model keyframe from parent to child node

kokosha:
The RSM model doesn't propagate keyframe animation information from parent nodes to child nodes. Need to pass the information properly from parent nodes to child nodes.
The RSM model is also separated in static model and dynamic model for faster calculation in static model.

### DIFF
--- a/korangar/src/loaders/model/mod.rs
+++ b/korangar/src/loaders/model/mod.rs
@@ -257,8 +257,14 @@ impl ModelLoader {
         )
     }
 
-    pub fn calculate_transformation_matrix(node: &mut Node, is_root: bool, bounding_box: AABB, parent_matrix: Matrix4<f32>) {
-        node.transform_matrix = match is_root {
+    pub fn calculate_transformation_matrix(
+        node: &mut Node,
+        is_root: bool,
+        bounding_box: AABB,
+        parent_matrix: Matrix4<f32>,
+        is_static: bool,
+    ) {
+        let transform_matrix = match is_root {
             true => {
                 let translation_matrix = Matrix4::from_translation(-Vector3::new(
                     bounding_box.center().x,
@@ -266,14 +272,24 @@ impl ModelLoader {
                     bounding_box.center().z,
                 ));
 
-                parent_matrix * translation_matrix * node.transform_matrix
+                translation_matrix * node.transform_matrix
             }
-            false => parent_matrix * node.transform_matrix,
+            false => node.transform_matrix,
         };
 
-        node.child_nodes
-            .iter_mut()
-            .for_each(|child_node| Self::calculate_transformation_matrix(child_node, false, bounding_box, node.transform_matrix));
+        node.transform_matrix = match is_static {
+            true => parent_matrix * transform_matrix,
+            false => transform_matrix,
+        };
+
+        node.child_nodes.iter_mut().for_each(|child_node| {
+            Self::calculate_transformation_matrix(child_node, false, bounding_box, node.transform_matrix, is_static);
+        });
+    }
+
+    // A model is static, if it doesn't have any animations.
+    pub fn is_static(node: &Node) -> bool {
+        node.rotation_keyframes.is_empty() && node.child_nodes.iter().all(Self::is_static)
     }
 
     pub fn collect_model_textures(&self, textures: &mut HashSet<String>, model_file: &str) {
@@ -386,15 +402,18 @@ impl ModelLoader {
             model_data.shade_type == 2,
             model_data.animation_length,
         );
+
         let mut root_nodes = vec![root_node];
+        let is_static = root_nodes.iter().all(Self::is_static);
 
         for root_node in root_nodes.iter_mut() {
-            Self::calculate_transformation_matrix(root_node, true, bounding_box, Matrix4::identity());
+            Self::calculate_transformation_matrix(root_node, true, bounding_box, Matrix4::identity(), is_static);
         }
 
         let model = Model::new(
             root_nodes,
             bounding_box,
+            is_static,
             #[cfg(feature = "debug")]
             model_data,
         );
--- a/korangar/src/world/model/mod.rs
+++ b/korangar/src/world/model/mod.rs
@@ -23,11 +23,21 @@ use crate::world::Camera;
 pub struct Model {
     pub root_nodes: Vec<Node>,
     pub bounding_box: AABB,
+    pub is_static: bool,
     #[cfg(feature = "debug")]
     pub model_data: ModelData,
 }
 
 impl Model {
+    pub fn get_model_matrix(&self, transform: &Transform) -> Matrix4<f32> {
+        let translation_matrix = Matrix4::from_translation(transform.position.to_vec());
+        let rotation_matrix = Matrix4::from_angle_z(-transform.rotation.z)
+            * Matrix4::from_angle_x(-transform.rotation.x)
+            * Matrix4::from_angle_y(transform.rotation.y);
+        let scale_matrix = Matrix4::from_nonuniform_scale(transform.scale.x, -transform.scale.y, transform.scale.z);
+        translation_matrix * rotation_matrix * scale_matrix
+    }
+
     pub fn render_geometry(
         &self,
         instructions: &mut Vec<ModelInstruction>,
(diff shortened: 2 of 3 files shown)


vE5li:
Can you add some more information to this PR please? I can't really tell what is being fixed here

kokosha:
The idea is to fix this rotation in #165.
I am thinking of improving this by doing some memoization, it is a draft.

vE5li:
If it's a draft it is helpful if you mark it as such :)

kokosha:
I think this is the best memoization that I can do with one matrix multiplication. To use no matrix multiplication would require
information from Model without Arc to be stored in the struct Object.

hasenbanck (changes requested):
I tested the functionality and it works great! But I have some comments about the implementation itself.

korangar/src/world/model/node.rs
```diff
@@ -49,41 +50,45 @@ impl Node {
...
-        let draw_oder_offset = (node_index as f32) * 1.1920929e-4_f32;
+        let draw_order_offset = (node_index as f32) * 1.1920929e-4_f32;
 
-        let model_matrix = self.world_matrix(transform, client_tick);
+        let model_matrix = self.world_matrix(client_tick, parent_matrix, is_static);
         let position = model_matrix.transform_point(self.centroid);
-        let distance = camera.distance_to(position) + draw_oder_offset;
+        let distance = camera.distance_to(position) + draw_order_offset;
+
+        // When is static, the parent matrix is only the shift from the object
```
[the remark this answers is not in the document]
kokosha: I changed the comment and wish it would be understandable

korangar/src/world/model/mod.rs
```diff
@@ -35,10 +37,16 @@ impl Model {
...
-            .enumerate()
-            .for_each(|(node_index, node)| node.render_geometry(instructions, transform, client_tick, camera, node_index));
+        self.root_nodes.iter().enumerate().for_each(|(node_index, node)| {
+            let translation_matrix = Matrix4::from_translation(transform.position.to_vec());
+            let rotation_matrix = Matrix4::from_angle_z(-transform.rotation.z)
+                * Matrix4::from_angle_x(-transform.rotation.x)
+                * Matrix4::from_angle_y(transform.rotation.y);
+            let scale_matrix = Matrix4::from_nonuniform_scale(transform.scale.x, -transform.scale.y, transform.scale.z);
+
+            let default_matrix = translation_matrix * rotation_matrix * scale_matrix;
```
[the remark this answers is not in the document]
kokosha: This is a matrix shift in world space for different objects but same models. I am thinking of naming world matrix.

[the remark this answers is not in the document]
hasenbanck: But that implies that is contains the translation inside the world space, meaning the map! Since it's the translation inside the model space, it's still the model_matrix.

korangar/src/world/model/node.rs
```diff
@@ -49,41 +50,45 @@ impl Node {
...
-        let animation_rotation_matrix = match self.rotation_keyframes.is_empty() {
-            true => Matrix4::identity(),
-            false => self.animaton_matrix(client_tick),
-        };
-
-        let rotation_matrix = Matrix4::from_angle_z(-transform.rotation.z)
-            * Matrix4::from_angle_x(-transform.rotation.x)
-            * Matrix4::from_angle_y(transform.rotation.y);
+    pub fn world_matrix(&self, client_tick: ClientTick, parent_matrix: Matrix4<f32>, is_static: bool) -> Matrix4<f32> {
+        if is_static == true {
```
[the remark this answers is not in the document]
hasenbanck: I don't see the problem:

It doesn't matter that the Model is inside an Arc! You don't need to change the transformation_matrix at runtime! You know if it's static while you are loading the model / node! You can decide if it's a static or dynamic matrix before you store the Model inside an Arc! There is nothing at runtime that could change the "is static" flag!

korangar/src/world/model/node.rs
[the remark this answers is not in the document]
hasenbanck: Maybe this is even better?

```
// When the render_geometry is set as static, the model matrix
// is already pre-calculated.
// When the render_geometry is set as dynamic, the model matrix
// needs to be calculated recursively, because the nodes from the model change
// positions due to animation motion.
```

korangar/src/loaders/model/mod.rs
```diff
@@ -266,14 +267,25 @@ impl ModelLoader {
...
-        node.child_nodes
-            .iter_mut()
-            .for_each(|child_node| Self::calculate_transformation_matrix(child_node, false, bounding_box, node.transform_matrix));
+        node.child_nodes.iter_mut().for_each(|child_node| {
+            Self::calculate_transformation_and_static_matrix(child_node, false, bounding_box, node.static_node_matrix);
+        });
+    }
+
+    // A model is static, if it doesn't have any animations.
+    pub fn is_static(node: &mut Node) -> bool {
```
vE5li, line 282: Why does this need to be mut? I don't see any mutation happening

korangar/src/loaders/model/mod.rs
```diff
@@ -389,12 +401,17 @@ impl ModelLoader {
         let mut root_nodes = vec![root_node];
 
         for root_node in root_nodes.iter_mut() {
-            Self::calculate_transformation_matrix(root_node, true, bounding_box, Matrix4::identity());
+            Self::calculate_transformation_and_static_matrix(root_node, true, bounding_box, Matrix4::identity());
         }
 
+        let is_static = root_nodes
+            .iter_mut()
+            .fold(true, |is_static, root_node| is_static & Self::is_static(root_node));
```
vE5li, line 409: If `is_static` doesn't mutate the state, it could be `any` instead of `fold`.

korangar/src/world/model/node.rs
```diff
@@ -49,41 +50,48 @@ impl Node {
...
-        let animation_rotation_matrix = match self.rotation_keyframes.is_empty() {
-            true => Matrix4::identity(),
-            false => self.animaton_matrix(client_tick),
-        };
-
-        let rotation_matrix = Matrix4::from_angle_z(-transform.rotation.z)
-            * Matrix4::from_angle_x(-transform.rotation.x)
-            * Matrix4::from_angle_y(transform.rotation.y);
+    pub fn world_matrix(&self, client_tick: ClientTick, parent_matrix: &Matrix4<f32>, is_static: bool) -> Matrix4<f32> {
+        if is_static == true {
```
vE5li, line 54: `if foo == true` is very much not idiomatic. It should just be `if foo`.

I am surprised that clippy is not complaining about this.

vE5li (changes requested):

korangar/src/loaders/model/mod.rs
kokosha: Copy pasta =/

korangar/src/loaders/model/mod.rs
kokosha: There is an even better one called `all`.

korangar/src/world/model/node.rs
kokosha: Now using match for this kind of comparison.

korangar/src/world/model/node.rs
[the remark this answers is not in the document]
kokosha: Not sure anymore, I was trying to implement, but there is a model cache that is returning the model directly as an Arc.

korangar/src/loaders/model/mod.rs
kokosha: Great catch!

korangar/src/loaders/model/mod.rs
```diff
@@ -257,23 +257,43 @@ impl ModelLoader {
...
+    }
+
+    // A model is static, if it doesn't have any animations.
+    pub fn is_static(node: &Node) -> bool {
+        let mut is_static = node.rotation_keyframes.is_empty();
+        node.child_nodes.iter().for_each(|child_node| {
+            let is_static_child = Self::is_static(child_node);
+            is_static &= is_static_child;
+        });
+        is_static
```
vE5li, line 296: This can be expressed like

```rust
        node.rotation_keyframes.is_empty() && node.child_nodes.iter().all(Self::is_static)
```

The advantages here are
- no `mut` required
- the function can exit early now as soon as something dynamic is found

vE5li (changes requested):

hasenbanck (approved):
LGTM now. But would need a squash before merge.

PR #265, opened 2025-02-24
Fix RSM model rotation animation length

kokosha:
The rotation animation length for some RSM model are wrong, for example the droplet of Izlude fountain.

### DIFF
--- a/korangar/src/world/model/node.rs
+++ b/korangar/src/world/model/node.rs
@@ -18,18 +18,24 @@ pub struct Node {
     pub vertex_offset: usize,
     pub vertex_count: usize,
     pub child_nodes: Vec<Node>,
+    pub animation_length: u32,
     pub rotation_keyframes: Vec<RotationKeyframeData>,
 }
 
 impl Node {
     fn animaton_matrix(&self, client_tick: ClientTick) -> Matrix4<f32> {
-        let last_step = self.rotation_keyframes.last().unwrap();
-        let animation_tick = client_tick.0 % last_step.frame;
+        let animation_tick = client_tick.0 % self.animation_length;
 
-        let mut last_keyframe_index = 0;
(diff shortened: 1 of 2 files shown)


korangar/src/world/model/node.rs
```diff
@@ -19,17 +19,35 @@ pub struct Node {
...
+        let animation_tick = client_tick.0 % self.animation_length;
 
         let mut last_keyframe_index = 0;
+
+        /* Old function
         while self.rotation_keyframes[last_keyframe_index + 1].frame < animation_tick {
             last_keyframe_index += 1;
+        }*/
+
+        // Binary search the exact index of keyframe
```
hasenbanck, line 37: Please remove the old code and use the Rust implementation of a binary search. My suggested version:

```rust
        let last_keyframe_index = self
            .rotation_keyframes
            .binary_search_by(|keyframe| keyframe.frame.cmp(&animation_tick))
            .unwrap_or_else(|keyframe_index| {
                // Err(i) returns the index where the searched element could bw insert to retain
                // the sort order. This means, that we haven't reached a new keyframe yet and
                // need to use the previous keyframe, hence the saturating sub.
                keyframe_index.saturating_sub(1)
            });
```

korangar/src/world/model/node.rs
```diff
@@ -19,17 +19,35 @@ pub struct Node {
     pub vertex_count: usize,
     pub child_nodes: Vec<Node>,
     pub rotation_keyframes: Vec<RotationKeyframeData>,
+    #[hidden_element]
```
hasenbanck, line 22: Is there a reason to hide it?

hasenbanck (changes requested):

hasenbanck:
Thanks for the PR. The rotation animation seem to properly work now.

hasenbanck:
LGTM. Will merge after I had something to eat!

hasenbanck (approved):

PR #269, opened 2025-03-02
Add static RSM2 model

kokosha:
This pull request add RSM2 model in static mode.

### DIFF
--- a/korangar/Cargo.toml
+++ b/korangar/Cargo.toml
@@ -37,6 +37,7 @@ rayon = { workspace = true }
 ron = { workspace = true }
 serde = { workspace = true }
 sevenz-rust2 = { workspace = true, features = ["compress", "zstd"] }
+smallvec = { workspace = true, features = ["const_generics", "const_new", "union"] }
 spin_sleep = { workspace = true }
 sys-locale = { workspace = true }
 walkdir = { workspace = true }
--- a/korangar/src/loaders/model/mod.rs
+++ b/korangar/src/loaders/model/mod.rs
@@ -1,17 +1,19 @@
 use std::sync::Arc;
 
-use cgmath::{EuclideanSpace, Matrix4, Point3, Rad, SquareMatrix, Vector2, Vector3};
+use cgmath::{Array, EuclideanSpace, Matrix4, Point3, Rad, SquareMatrix, Vector2, Vector3};
 use derive_new::new;
-use hashbrown::HashSet;
+use hashbrown::{HashMap, HashSet};
 #[cfg(feature = "debug")]
 use korangar_debug::logging::{Colorize, Timer, print_debug};
 use korangar_util::FileLoader;
 use korangar_util::collision::AABB;
 use korangar_util::math::multiply_matrix4_and_point3;
 use korangar_util::texture_atlas::AllocationId;
+use num::Zero;
 use ragnarok_bytes::{ByteReader, FromBytes};
 use ragnarok_formats::model::{ModelData, NodeData};
 use ragnarok_formats::version::InternalVersion;
+use smallvec::{SmallVec, smallvec};
 
 use super::error::LoadError;
 use super::{FALLBACK_MODEL_FILE, smooth_model_normals};
@@ -30,7 +32,7 @@ impl ModelLoader {
         vertices: &mut [NativeModelVertex],
         vertex_positions: &[Point3<f32>],
         texture_coordinates: &[Vector2<f32>],
-        smoothing_groups: &[i32; 3],
+        smoothing_groups: &SmallVec<[i32; 3]>,
         texture_index: i32,
         reverse_vertices: bool,
         reverse_normal: bool,
@@ -54,7 +56,7 @@ impl ModelLoader {
                     texture_index,
                     Color::WHITE,
                     0.0, // TODO: actually add wind affinity
-                    *smoothing_groups,
+                    smoothing_groups.clone(),
                 );
             }
         } else {
@@ -68,7 +70,7 @@ impl ModelLoader {
                     texture_index,
                     Color::WHITE,
                     0.0, // TODO: actually add wind affinity
-                    *smoothing_groups,
+                    smoothing_groups.clone(),
                 );
             }
         }
@@ -84,7 +86,7 @@ impl ModelLoader {
         let mut face_index = 0;
         let mut back_face_index = face_vertex_count;
 
-        let array: [f32; 3] = node.scale.unwrap().into();
+        let array: [f32; 3] = node.scale.unwrap_or(Vector3::new(1.0, 1.0, 1.0)).into();
         let reverse_node_order = array.into_iter().fold(1.0, |a, b| a * b).is_sign_negative();
 
         if reverse_node_order {
@@ -102,14 +104,12 @@ impl ModelLoader {
                 let coordinate_index = face.texture_coordinate_indices[index];
                 node.texture_coordinates[coordinate_index as usize].coordinates
             });
-
-            let smoothing_groups = match face.smooth_group_extra.as_ref() {
-                None => [face.smooth_group, -1, -1],
-                Some(extras) if extras.len() == 1 => [face.smooth_group, extras[0], -1],
-                Some(extras) if extras.len() == 2 => [face.smooth_group, extras[0], extras[1]],
-                _ => panic!("more than three smoothing groups found"),
+            let mut smoothing_groups: SmallVec<[i32; 3]> = smallvec![face.smooth_group];
+            if let Some(extras) = face.smooth_group_extra.as_ref() {
+                for extra in extras {
+                    smoothing_groups.push(*extra);
+                }
             };
-
             Self::add_vertices(
                 &mut native_vertices[face_index..face_index + 3],
                 &vertex_positions,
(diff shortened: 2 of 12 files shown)


korangar/src/loaders/model/mod.rs
```diff
@@ -107,7 +107,8 @@ impl ModelLoader {
                 None => [face.smooth_group, -1, -1],
                 Some(extras) if extras.len() == 1 => [face.smooth_group, extras[0], -1],
                 Some(extras) if extras.len() == 2 => [face.smooth_group, extras[0], extras[1]],
-                _ => panic!("more than three smoothing groups found"),
+                // TODO: add for more than two smoothing groups, default as no smoothing group
```
hasenbanck, line 110: We could use "SmallVec" in this case, where we optimize for 3 smooth groups and allow to grow more.

korangar/src/loaders/model/mod.rs
```diff
@@ -144,21 +145,35 @@ impl ModelLoader {
...
+        match version.equals_or_above(2, 2) {
+            true => {
+                let main = Matrix4::from(node.offset_matrix);
+                let translation_matrix = Matrix4::from_translation(node.translation2);
+                let transform = translation_matrix;
+                (main, transform, transform)
+            }
+            false => {
+                let main = Matrix4::from_translation(node.translation1.unwrap()) * Matrix4::from(node.offset_matrix);
+                let scale = node.scale.unwrap();
```
hasenbanck, line 162: Can we also just use an identity vector instead? (even if it doesn't occur)

korangar/src/loaders/model/mod.rs
```diff
@@ -169,19 +184,21 @@ impl ModelLoader {
...
 
     fn process_node_mesh(
+        version: InternalVersion,
         current_node: &NodeData,
         nodes: &[NodeData],
         processed_node_indices: &mut [bool],
         vertex_offset: &mut usize,
         native_vertices: &mut Vec<NativeModelVertex>,
         model_texture_mapping: &[ModelTexture],
+        hashmap_texture: &HashMap<String, i32>,
```
hasenbanck, line 194: Can't we use an enum instead? We could then just use one of the texture mappings (vec or map)

```rust
enum TextureMapping<'a> {
    PreVersion2_3(&'a[ModelTexture])
    PostVersion2_3(&'aHashMap<String, i32>)
}
````

File: korangar/src/loaders/model/mod.rs [hunk 4]
```diff
@@ -359,12 +396,31 @@ impl ModelLoader {
...
+            false => model_data
+                .texture_names
+                .iter()
+                .map(|texture_name| texture_name.as_ref().to_string())
+                .collect(),
+            true => model_data
+                .nodes
+                .iter()
+                .flat_map(|node_data| node_data.texture_names.iter().map(|name| name.as_ref().to_string()))
+                .collect::<HashSet<_>>() // TODO: seems not deterministic
```
hasenbanck, line 409: It should have a concrete type.

korangar/src/loaders/model/mod.rs
```diff
@@ -359,12 +396,31 @@ impl ModelLoader {
...
+                .into_iter()
+                .collect(),
+        };
+
+        let texture_allocation: Vec<TextureAtlasEntry> = texture_names
             .iter()
             .map(|texture_name| texture_atlas.register(texture_name.as_ref()))
             .collect();
 
+        let mut hashmap_texture = HashMap::<String, i32>::new();
```
hasenbanck, line 419: I think you can use `HashMap::from_iter()` here.

korangar/src/loaders/model/mod.rs
```diff
@@ -374,13 +430,22 @@ impl ModelLoader {
             })
             .collect();
 
-        let root_node_name = &model_data.root_node_name.clone().unwrap();
+        let mut root_node_names = Vec::<ModelString<40>>::new();
 
+        if version.equals_or_above(2, 2) {
```
hasenbanck, line 435: Do we also want to use a map here? In the other places with the version we also did so.

korangar/src/loaders/model/mod.rs
```diff
@@ -374,13 +430,22 @@ impl ModelLoader {
...
+            for i in 0..model_data.root_node_count.unwrap() {
+                root_node_names.push(model_data.root_node_names[i as usize].clone());
+            }
+        } else {
+            let root_node_name = model_data.root_node_name.clone().unwrap();
+            root_node_names.push(root_node_name);
+        }
+        // TODO: CHECK FOR OTHER ROOT NODES
```
hasenbanck, line 443: Do you want to do this in this PR? Our model structure should allow for multiple root nodes already.

korangar/src/loaders/model/mod.rs
```diff
@@ -390,27 +455,36 @@ impl ModelLoader {
...
             model_data.shade_type == 2,
             model_data.animation_length,
         );
 
+        drop(hashmap_texture);
+
         let mut root_nodes = vec![root_node];
-        let is_static = root_nodes.iter().all(Self::is_static);
+        let is_static = match version.equals_or_above(2, 2) {
+            // TODO: Need to add the dynamic RSM2.
```
hasenbanck, line 477: Is this TODO still open?

korangar/src/world/model/mod.rs
```diff
@@ -34,7 +36,10 @@ impl Model {
         let rotation_matrix = Matrix4::from_angle_z(-transform.rotation.z)
             * Matrix4::from_angle_x(-transform.rotation.x)
             * Matrix4::from_angle_y(transform.rotation.y);
-        let scale_matrix = Matrix4::from_nonuniform_scale(transform.scale.x, -transform.scale.y, transform.scale.z);
+        let scale_matrix = match self.version.equals_or_above(2, 2) {
```
hasenbanck, line 39: Did they really flip the scale dimension direction? .... oh gravity ...

hasenbanck (changes requested):

korangar/src/world/model/mod.rs
kokosha: Not sure, when I loaded everything was upside down, so I just turned the negative sign and it worked after removing the bounding box shift. I  think is a fix from the original method that was already upside down.

korangar/src/loaders/model/mod.rs
kokosha: It is still open, I am trying to make the dynamic work.

korangar/src/loaders/model/mod.rs
kokosha: I was thinking of it, but doesn't find any model with multiple root nodes. So I postponed.

korangar/Cargo.toml
```diff
@@ -37,6 +37,7 @@ rayon = { workspace = true }
 ron = { workspace = true }
 serde = { workspace = true }
 sevenz-rust2 = { workspace = true, features = ["compress", "zstd"] }
+smallvec =  { workspace = true }
```
hasenbanck, line 40:
```suggestion
smallvec = { workspace = true, features = ["const_generics", "const_new", "union"] }
```

This makes SmallVec simpler to use and more efficient. You also have access to "const_new()" then.

korangar/src/loaders/model/mod.rs
```diff
@@ -39,6 +41,8 @@ impl ModelLoader {
             true => NativeModelVertex::calculate_normal(vertex_positions[0], vertex_positions[1], vertex_positions[2]),
             false => NativeModelVertex::calculate_normal(vertex_positions[2], vertex_positions[1], vertex_positions[0]),
         };
+        // TODO: more than 3 smoothing group isn't handled.
```
hasenbanck, line 44: Let's properly support the case when we have more then 3 smooth groups. So you need to remove this and instead make change the NativeVertex to use SmallVec. You can use "const_new()" then. It's fine it NativeVertex loses it's copy propery and instead only implements clone. Smoothing groups should keep working as they are now, since they don't have the "-1" handled anywhere.

korangar/src/loaders/model/mod.rs
```diff
@@ -102,14 +106,15 @@ impl ModelLoader {
...
-                Some(extras) if extras.len() == 2 => [face.smooth_group, extras[0], extras[1]],
-                _ => panic!("more than three smoothing groups found"),
+            let mut smoothing_groups: SmallVec<[i32; 3]> = smallvec![face.smooth_group];
+            if let Some(extras) = face.smooth_group_extra.as_ref() {
+                for extra in extras {
+                    smoothing_groups.push(*extra);
+                }
             };
-
+            if smoothing_groups.len() < 3 {
```
hasenbanck, line 115: You can also remove this.

korangar/src/loaders/model/mod.rs
```diff
@@ -205,12 +214,13 @@ impl ModelLoader {
...
             .map(|&index| {
                 Self::process_node_mesh(
+                    version,
                     &nodes[index],
                     nodes,
                     processed_node_indices,
                     vertex_offset,
                     native_vertices,
-                    model_texture_mapping,
+                    &texture_mapping,
```
hasenbanck, line 223: You can remove the & here.

korangar/src/loaders/model/mod.rs
```diff
@@ -359,60 +385,122 @@ impl ModelLoader {
...
+                        transparent: entry.transparent,
+                    })
+                    .collect();
+                TextureMapping::PreVersion2_3(vector_texture)
+            }
+        };
 
-        let root_node_name = &model_data.root_node_name.clone().unwrap();
+        let root_node_names = match version.equals_or_above(2, 2) {
+            true => model_data.root_node_names.iter().map(|name| name.clone()).collect(),
```
hasenbanck, line 438:
```suggestion
            true => model_data.root_node_names.to_vec(),
```

korangar/src/loaders/model/mod.rs
```diff
@@ -359,60 +385,122 @@ impl ModelLoader {
...
+            })
+            .unzip();
 
+        drop(texture_mapping);
+        let is_static = match version.equals_or_above(2, 2) {
+            // TODO: Need to add the dynamic RSM2.
+            true => true,
+            false => root_nodes.iter().all(Self::is_static),
+        };
+        let bounding_box = bounding_boxes
```
hasenbanck, line 489: Instead of doing it this way, just pull the `let mut bounding_box = AABB::uninitialized();` outside of the loop. The bounding box get's extended inside the `process_node_mesh()` anyhow.

hasenbanck (changes requested):

korangar/src/loaders/model/mod.rs
kokosha: Sure, but I am not finished with dynamic models yet

korangar/src/loaders/model/mod.rs
```diff
@@ -271,25 +349,68 @@ impl ModelLoader {
...
-        node.rotation_keyframes.is_empty() && node.child_nodes.iter().all(Self::is_static)
+        node.scale_keyframes.is_empty()
+            && node.translation_keyframes.is_empty()
+            && node.rotation_keyframes.is_empty()
+            && node.child_nodes.iter().all(Self::is_static)
+    }
+
+    /// The texture names should remains equal to generate the texture atlas
+    /// correctly.
+    pub fn collect_versioned_texture_names(version: &InternalVersion, model_data: &ModelData) -> Vec<String> {
```
hasenbanck, line 392: Function doesn't need to be pub!

I would also change the comment like this:

```rust
    /// We need to make sure to always generate a texture atlas in the same
    /// order when creating an online texture atlas and an offline texture
    /// atlas.
    fn collect_versioned_texture_names(version: &InternalVersion, model_data: &ModelData) -> Vec<String> {
```

hasenbanck (changes requested):

hasenbanck (approved):

PR #271, opened 2025-03-09
Add the build number to the map loader

kokosha:
Some maps need to use build number to parse the information for example 1@exds.

### DIFF
--- a/ragnarok_procedural/src/utils.rs
+++ b/ragnarok_procedural/src/utils.rs
@@ -31,6 +31,20 @@ impl Parse for Version {
     }
 }
 
+#[derive(Clone)]
+pub struct VersionBuild {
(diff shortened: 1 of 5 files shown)


ragnarok_procedural/src/utils.rs
```diff
@@ -31,6 +31,20 @@ impl Parse for Version {
     }
 }
 
+#[derive(Clone)]
+pub struct VersionBuild {
+    pub version: Version,
+    pub build: LitInt,
+}
+impl Parse for VersionBuild {
```
hasenbanck, line 39: Small nit: Can you include aspace between the struct and impl?

hasenbanck (approved):
LGTM

PR #302, opened 2025-07-07
Update nightly version of Rust

hasenbanck:
Some unstable stuff turned stable. For negative traits we had to use the allocator API to be able to properly define the trait.

Also fixed clippy issues that popped up.

@vE5li Some stuff that might interact with your branch.


hasenbanck:
Closed for now until the UI branch is merged.

Issue #306, opened 2025-08-18
Multi-Version Game Support


nmeylan:
# Multi-Version Game Support

## Overview

The goal of this issue is to discuss with the korangar team if supporting multiple versions of the game:

1. Could be possible within korangar
1. Or, korangar could be designed in a way to allow support of different game versions via extension (fork, plugin, dependency injection)
1. Or, is unwanted by the korangar team

## Problem Statement

Currently, Korangar supports only one game version, which limits its utility despite the significant development effort invested. 

Supporting multiple versions is not an easy task as it**requires handling divergent**:

- **Packet structures**
- **User interface components**
- **Game resources**

Some features exist only in newer versions (e.g., Rodex), while others have evolved significantly (equipment window merged with status panel in recent versions vs. separate panels in older versions).

## Technical Implementation Analysis
Servers implementation or at least one other client (robrowser) support multiple version of the game, in section below I will try to analyse how this is achieved, to set bases for a discussion.

### 1. Multi-Version Packet Support

Packet evolution occurs in two primary patterns:

#### Pattern 1: Field Updates in Existing Packets
Example: `PACKET_HC_ACCEPT_ENTER_NEO_UNION(0x006b)`
- Packet version ≥ `20170830`: `exp` is `int64`
- Packet version < `20170830`: `exp` is `int32`
- Packet version ≥ `20141022`: `body` field added

#### Pattern 2: Packet Replacement
Example: `PACKET_CH_MAKE_CHAR(0x0067)` → `PACKET_CH_MAKE_CHAR2(0x0970)`

#### Option A: Compile-Time Definition (rAthena/Hercules Style)
Server implementations use compilation-time packet structure modification:

```c
// Packet structure includes all field definitions with version conditions
#if PACKETVER < 20110817
    // older field definitions
#endif
```

**Advantages:**
- Proven approach in production servers
- Clear version-specific code paths

**Disadvantages:**
- Code cluttered with `#if PACKETVER` conditionals
- Cannot test multiple packet versions without recompilation

#### Option B: Runtime Support (rust-ro Style)
Packet structures contain all possible fields with runtime version handling.

Example from [rust-ro CharacterInfoNeoUnion](https://github.com/nmeylan/rust-ro/blob/e080cc2c8ef8fb14a4cbb963ff410717b22268a3/lib/packets/src/packets.rs#L10524):

```rust
impl CharacterInfoNeoUnion {
[...]

PR #311, opened 2025-08-25
Proposal for different packet versions handling.

nmeylan:
This is a proposal for #306 
This proposal provides to korangar capability to handle different packets versions, leveraging existing handler mecanism.
For packet emission, a new `factory` concept has been added.

# Goals
Goals achieved by this proposal:
- Korangar is unaware of packet version
- There at the moment no condition on packet version
- Korangar is now extensible by any 3rd party developer who want to use its own packet version. Support of a specific version of packet can be done in an external crate, one can maintain its own crate of packet handling and inject its implementation in korangar main without any changes.
- Minimize maintenance effort in korangar 
- Avoid impacting korangar compile time
- Avoid impacting korangar binary size
- For now there is no feature flags, customization is achieved via dependency injection

# Packet Factory
The concept of factory is to map game data into packet ready to be sent to the wire.
Factory are traits to be implemented to emit packet.

The "downside" of this approach is that between 2 specific versions, factory code will duplicated at a high percentage depending on the versions gap, but IMO it is not an issue.

# Questions

## Packet handler
Should packet handler be registered for a specific version only?

In my proposal I kept handling of existing packets as it was in `NetworkingSystem`, with a mecanism to inject a `register` which can be used to register additional handler.
With the approach describe above we can encounter `DuplicateHandlerError`, this was the case with `CharacterListPacket` which had a `noop` handler register, but in my example i want to handle this packet as it is sent by the server in my specific version.

Thus i am not sure if I should go further in the approach and instead of allowing registering additionnal packet handler, packet handler should be defined for each versions (as for factory) and finding something to share handler between versions.

## Injection
In my commit the injection of packet factory and handler register is done in korangar main.
What would be cool would be to expose `Client` structure as public in `korangar` crate, so a 3rd party developer would have only few line of code to use it manage it own version of the game. 

# Next step
I would like to implement packets for the version supported by rust-ro, I was thinking to do that in a dedicated crate "korangar-packet-20120307" (on my own space?).
I don't know if packets should be added in `ragnarok-packets` crate, as it may slow down compilation time, is it an issue? I think this is the place for packets

# Side notes
Existing design and code was a pleasure to work with, I like the way networking and packets were implemented.
The packet handler mecanism is more elegant than the one I have implemented in rust-ro.

### DIFF
--- a/korangar-networking/Cargo.toml
+++ b/korangar-networking/Cargo.toml
@@ -12,9 +12,13 @@ tokio = { workspace = true, features = ["rt", "io-util", "net", "macros", "sync"
 
 [dev-dependencies]
 korangar-debug = { workspace = true }
+ragnarok-packets = { workspace = true, features = ["derive"]}
 reqwest = { workspace = true, features = ["json"] }
 serde = { workspace = true, features = ["derive"] }
 
 [features]
 debug = []
-interface = ["korangar-interface", "rust-state"]
+interface = ["korangar-interface", "rust-state", "ragnarok-packets/packet-to-state-element"]
+# This feature is required for specific-packet-version example, without it when running `cargo test` or `cargo test --all-features` on workspace
+# compiler will complain "missing `to_element` in implementation".
+packet-to-state-element = []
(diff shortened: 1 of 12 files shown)


vE5li:
See https://github.com/vE5li/korangar/pull/328

nmeylan:
I am closing this PR, as we have discussed in discord, this approach bring too much abstraction

PR #313, opened 2025-08-25
Improve user experience

vE5li:
This PR does three things:
- Automatically adjust the working directory from the root of the repository to the `korangar` subdirectory if no `archive` directory is found
- In case an archive is not found, show the file name and a custom error message 
- Show errors instead of crashing if settings or language files can not be written


hasenbanck (approved):
LGTM

PR #353, opened 2025-09-27
Remove `korangar-util` crate

vE5li:
I feel like `util` modules and crates often lead to mixing of domains. This PR splits the `korangar-util` crate into multiple crates, separated by their use case.

This also prepares some future rework, such as creating the `korangar-loaders` crate that I want to eventually move the loaders to.


vE5li:
@hasenbanck I know that some of the structs/modules are not in the ideal location yet but I think we can just improve over time. At least for me this is already quite a bit nicer than before.

hasenbanck:
LGTM. The graphics module is also a candidate for me to move into it's own crate sooner than later.

hasenbanck (approved):

vE5li:
Same, also because we can't separate the loaders without having the graphics in a sub crate.

PR #359, opened 2025-10-05
WIP: Implementing new skill tree

vE5li:


korangar/src/interface/windows/skill_tree.rs
```diff
@@ -1,27 +1,163 @@
...
+{
+    fn follow<'a>(&self, state: &'a ClientState) -> Option<&'a LearnableSkill> {
+        // SAFETY:
+        // Unwrapping is save because of the bounds
+        self.layout_path.follow(state).unwrap().get(&self.index)
+    }
+
+    fn follow_mut<'a>(&self, state: &'a mut ClientState) -> Option<&'a mut LearnableSkill> {
+        // SAFETY:
+        // Unwrapping is save because of the bounds
```
rdw-software, line 55:
```suggestion
        // Unwrapping is safe because of the bounds
```

korangar/src/interface/windows/skill_tree.rs
```diff
@@ -1,27 +1,163 @@
...
+    }
+}
+
+impl<A> Path<ClientState, LearnableSkill, false> for LearnableSkillPath<A>
+where
+    A: Path<ClientState, HashMap<usize, LearnableSkill>>,
+{
+    fn follow<'a>(&self, state: &'a ClientState) -> Option<&'a LearnableSkill> {
+        // SAFETY:
+        // Unwrapping is save because of the bounds
```
rdw-software, line 49:
```suggestion
        // Unwrapping is safe because of the bounds
```

(Note: Not really clear *why* to an outside observer/Rust dummy, but that's besides the point)

korangar/src/inventory/skills.rs
```diff
@@ -1,17 +1,91 @@
...
+where
+    A: Path<ClientState, LearnableSkill, false>,
+    B: Path<ClientState, Vec<LearnedSkill>>,
+{
+    fn follow<'a>(&self, state: &'a ClientState) -> Option<&'a LearnedSkill> {
+        let learnable_skill = self.learnable_skill_path.follow(state)?;
+        let learnable_skill_id = learnable_skill.skill_id;
+
+        // SAFETY:
+        // Unwrap is safe here because of the bounds.
```
rdw-software, line 56: Seeing how this is repeated 4 times, maybe an inlined utility would be a better place to store it? (if not too costly)

korangar/src/main.rs
```diff
@@ -1669,18 +1680,20 @@ impl Client {
...
-                                    .find_skill(SkillId(hotkey.skill_id as u16))
-                                else {
-                                    self.client_state
-                                        .follow_mut(client_state().hotbar())
-                                        .clear_slot(&mut self.networking_system, HotbarSlot(index as u16));
-                                    continue;
-                                };
-
-                                skill.skill_level = hotkey.quantity_or_skill_level;
+                                // TODO: Properly distinct between skill and item.
```
rdw-software, line 1683:
```suggestion
                                // TODO: Properly distinguish between skill and item.
```

korangar/src/world/entity/mod.rs
```diff
@@ -100,9 +102,9 @@ pub enum EntityType {
...
 }
 
-impl From<usize> for EntityType {
-    fn from(value: usize) -> Self {
-        match value {
+impl From<JobId> for EntityType {
+    fn from(job_id: JobId) -> Self {
+        match job_id.0 {
             45 => EntityType::Warp,
             111 => EntityType::Hidden, // TODO: check that this is correct
```
rdw-software, line 109: Side note (not touched here): According to the [Hercules docs](https://github.com/HerculesWS/Hercules/blob/c1fa5502eb08143a0e6e7862577fa57480919c0f/doc/constants_pre-re.md?plain=1#L2527) this is indeed correct.

korangar/src/world/library/mod.rs
```diff
@@ -164,7 +197,98 @@ impl Library {
...
+                    Ok(table) => {
+                        let mut job_specific = HashMap::new();
+
+                        for (job_id, sequence) in table.pairs::<u16, mlua::Table>().flatten() {
+                            let mut required_skills = HashMap::new();
+
+                            for required_skill in sequence.sequence_values::<mlua::Table>() {
+                                let required_skill = required_skill?;
+                                let skill_id = required_skill.get::<u16>(1)?;
+                                // TODO: At least one skill does not have a level set. I am just
```
rdw-software, line 239: Feel free to disregard if this isn't intended to remain in the code: I'd always add which skill (example/ID/name) so that others don't need to search/run the code/wonder if the comment is outdated should they encounter it later.

(Even better, if maybe excessive: Programmatically encode assumptions and let the machine detect invalid ones)

PR #360, opened 2025-10-10
Port forward shader

hasenbanck:
Let's merge this before my PR gets merged in WGPU upstream.


hasenbanck:
Closed because #361 will superseed it.

PR #396, opened 2026-04-06
Add a basic minimap subsystem

septavios:
## Summary

- adds an initial in-game minimap implementation
- splits minimap logic into dedicated modules for generation, projection, markers, and rendering
- stores minimap state separately from the window implementation
- opens the minimap automatically in-game and allows toggling it from the UI
- uses a generated fallback texture derived from map tile data
- renders the local player and visible entities on the minimap
- supports zooming in the minimap window

## Notes

- this implementation uses generated map imagery
- marker rendering is still basic
- no quest, NPC, or warp overlay data yet
- no packet-driven minimap markers yet

Testing:
- Tested on mac m3 pro.

In game screenshot
image
image

## Validation

- `cargo check --release`
- `cargo clippy --release --no-deps`

### DIFF
--- a/korangar/shaders/modules/interface.slang
+++ b/korangar/shaders/modules/interface.slang
@@ -21,6 +21,8 @@ public struct RectangleInstanceData {
     public var texture_size: float2;
     public var rectangle_type: uint;
     public var texture_index: int;
+    public var rotation: float;
+    public var padding: uint;
 };
 
 public struct RectangleVertexInput {
@@ -258,7 +260,19 @@ public func rectangle_vertex_shader(
     adjusted_size += shadow_expansion;
 
     let clip_size = adjusted_size * 2.0;
-    let position = coordinate_space::screen_to_clip_space(instance.screen_position + position_offset) + vertex.xy * clip_size;
+
+    var rotated_vertex = vertex.xy * clip_size;
+    if (instance.rotation != 0.0) {
+        let center = float2(0.5, -0.5) * clip_size;
+        let c = cos(instance.rotation);
+        let s = sin(instance.rotation);
+        let p = rotated_vertex - center;
+        let rx = p.x * c - p.y * s;
+        let ry = p.x * s + p.y * c;
+        rotated_vertex = float2(rx, ry) + center;
+    }
+
+    let position = coordinate_space::screen_to_clip_space(instance.screen_position + position_offset) + rotated_vertex;
 
     var output: RectangleVertexOutput;
     output.position = float4(position, 0.0, 1.0);
--- a/korangar/shaders/modules/postprocessing.slang
+++ b/korangar/shaders/modules/postprocessing.slang
@@ -46,6 +46,8 @@ public struct RectangleInstanceData {
     public var texture_size: float2;
     public var rectangle_type: uint;
     public var texture_index: int;
+    public var rotation: float;
+    public var padding: uint;
 }
 
 public struct DebugUniforms {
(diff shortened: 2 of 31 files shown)


eleriaqueen:
Hi, I tested the PR in game and player direction arrow/indicator on the map is flipped.
Edit: The minimap itself might be flipped ? (See this screenshoot where minimap and actual buildings don't match at all)
Screenshot_20260407_141545

septavios:
> Hi, I tested the PR in game and player direction arrow/indicator on the map is flipped. Edit: The minimap itself might be flipped ? (See this screenshoot where minimap and actual buildings don't match at all) Screenshot_20260407_141545

image
image

Hi, thanks for testing, ya this is a bug. will fix it.
attached is my local test with the new fix.

septavios (commented):
56bcce61 (“Fix minimap generated texture Y-axis flip”)

septavios (commented):
Pushed a fix for the minimap flip: the generated minimap texture Y axis was inverted vs tile/world coordinates. New commit: 56bcce61. Please retest: minimap orientation + player arrow should now match the world.

septavios:
image

- Improved Player Marker (Dynamic Arrow):
- Replaced the text-based arrow fallback with a crisp, dynamically rendering SDF texture ( arrow_right.png ).
- Adjusted the SDF rendering bounds to stretch the arrow into a sharp, clear pointer shape.
- Re-colored the player marker to Bright Yellow for high contrast against the minimap terrain.

eleriaqueen:
Can't compile.
Screenshot_20260408_043127

Changing code to this (adding `rotation: 0.0,`) gets it to compile, but the "arrow" shape does not loog good, it is a bit too big, and does not point in the player's direction (it points in a slightly off direction):
```rust
 self.instructions.borrow_mut().push(RectangleInstruction::Sdf {
          screen_position,
          screen_size,
          color,
          texture_position: Vector2::new(0.0, 0.0),
          texture_size: Vector2::new(1.0, 1.0),
          texture: texture.clone(),
          rotation: 0.0,
});
```
Screenshot_20260408_045127

Otherwise the map part of the minimap looks fine now. It is no longer flipped.
Thank you for your efforts so far.

septavios:
Fixed in the latest commit.

Summary:

- Added the missing rotation field for RectangleInstruction::Sdf, which resolves the compile error.
- Removed the minimap arrow’s incorrect camera-based rotation adjustment.
- Corrected the player arrow rotation so it points in the proper direction on the now-unflipped minimap.
- Tuned the arrow size/shape and added a darker arrowhead overlay to make the tip direction clearer.
- Please test again when you get a chance — especially cardinal directions and diagonals on the minimap.


image

image
image

eleriaqueen:
Hey thanks for the update and for your work. 
I'm not the maintainer of the repo btw so take my mind with a grain of salt but I'm gonna be honest and say the representation of the character's orientation (the "arrow") in your PR currently doesn't look good. 
Its orientation is still a bit confusing (It doesn't look like a "real" rotation of your triangle shape ?), I don't think I know the right terms neither in English nor in my native language to explain the issue(s).

I would appreciate if someone else could maybe chime-in and give their thoughts ?

That being said:
The very first PRs might have looked slightly funny because you were using unicode characters or something but they had one key benefit which was that the shape of that arrow looked "correct".
It wasn't visible enough, not thick enough, maybe also not the right colour but when I managed to see the indicator on the minimap it was clear as day where it pointed.

Here is an ingame screenshot for people who don't want / can't run Korangar with the changes.

Screenshot_20260408_215931

eleriaqueen:
I threw this together quickly to try and hopefully illustrate the confusion of my brain trying to "read" the direction indicator. This might also be just a "me problem", again take with a grain of salt.
korangar-direction-indicator-confusion

septavios:
> I threw this together quickly to try and hopefully illustrate the confusion of my brain trying to "read" the direction indicator. This might also be just a "me problem", again take with a grain of salt. korangar-direction-indicator-confusion

The dark spot means its heading for Direction 2.
if the currently direction implementation is not that useful, maybe can just a round icon just to indicate the player position minimap.

Alternative is revert to the arrow using uncode if its better.

PR #397, opened 2026-04-06
Improve overly dark in-game lighting

septavios:
# Improve overly dark in-game lighting (configurable)

This PR adds a user-facing Brightness control to address maps that still appear too dark on some setups without hardcoding global lighting floors.

## What changed

- Add Graphics Settings → Brightness (50%–200%)
- Apply brightness in the final screen blit (affects final presented frame)
- Wire brightness through `Uniforms` → `GlobalUniforms` and fix uniform padding to satisfy wgpu validation (macOS/Metal)
- Bind the global bind group in the screen blit pass

## Why

Some maps still read too dark depending on renderer calibration / platform output, and hardcoded map-light normalization can be too aggressive.
A brightness control provides an immediate, user-adjustable workaround while the underlying lighting calibration is investigated.

## Scope

This PR touches:
- Graphics settings UI + persistence
- Global uniforms + screen blit shaders/pass

It no longer includes:
- Forcing an sRGB swapchain format
- Map light clamping / floor normalization

## Testing

- `cargo run --release` (macOS)
- Verified the Brightness setting changes output and persists via `client/graphics_settings.ron`

Before fix:

image

After fix:
image
image
image
image

## Potential follow-up

- Verify behavior on Windows and Linux
- Revisit whether some of these lighting floors should become configurable
- Re-evaluate map-specific lighting once broader renderer calibration is in place


hasenbanck:
Hi, thank you for your effort. But this PR is chasing two phantoms:

1. The sRGB change doesn't help in any way. We use the surface that the backend recommend, that can either be a sRGB or non-sRGB format. We compensate for non-sRGB formats in the [blitter shader](https://github.com/vE5li/korangar/blob/main/korangar/shaders/passes/screen_blit/blitter.slang).
2. The lightning problem you observe is not a problem with the lightning in general, but you observe a bug that we have on MacOS since 33be2c4a388f78db175a099394171a6e7e18c32a. You can test 5c2ea1ea0607f6114e0f9ed36a9474ec0d7e9eb5 if you want. The lightning is fixed in that commit.

We currently don't know why WGPU v28 (and v29) broke the lightning for MacOS. If you want to help, you could bisect WGPU and find out, which commit did break the lightning for us. Then we could either provide a workaround in our shaders, or a patch to WGPU to fix this issue properly.

septavios (commented):
Update codes

septavios:
> Hi, thank you for your effort. But this PR is chasing two phantoms:
> 
> 1. The sRGB change doesn't help in any way. We use the surface that the backend recommend, that can either be a sRGB or non-sRGB format. We compensate for non-sRGB formats in the [blitter shader](https://github.com/vE5li/korangar/blob/main/korangar/shaders/passes/screen_blit/blitter.slang).
> 2. The lightning problem you observe is not a problem with the lightning in general, but you observe a bug that we have on MacOS since [33be2c4](https://github.com/vE5li/korangar/commit/33be2c4a388f78db175a099394171a6e7e18c32a). You can test [5c2ea1e](https://github.com/vE5li/korangar/commit/5c2ea1ea0607f6114e0f9ed36a9474ec0d7e9eb5) if you want. The lightning is fixed in that commit.
> 
> We currently don't know why WGPU v28 (and v29) broke the lightning for MacOS. If you want to help, you could bisect WGPU and find out, which commit did break the lightning for us. Then we could either provide a workaround in our shaders, or a patch to WGPU to fix this issue properly.

Thanks! Confirmed: Korangar already compensates non-sRGB surfaces in the screen-blit shader, so the sRGB changes aren’t useful here. Also confirmed the macOS ‘dark lighting’ regression starts at 33be2c4 (WGPU v28 bump) and is fixed by 5c2ea1e (forcing Slang to emit SPIR-V 1.3). I’ll rebase/remove the sRGB changes, and look into WGPU.

